PR #266, opened 2026-04-09
fix(auth): use SimpleUrlAuthenticationSuccessHandler for OAuth2 login

myml:
## Summary

- What changed?

Replace SavedRequestAwareAuthenticationSuccessHandler with SimpleUrlAuthenticationSuccessHandler to prevent redirecting to saved API requests after OAuth2 login.

- Why is this needed?

Previously, when a user accessed a protected API endpoint (e.g., /api/web/skills) without authentication, Spring Security would save that request. After OAuth2 login, the handler would redirect back to the API endpoint instead of the dashboard.

Now the handler only uses:
- returnTo parameter from session (if present)
- default target URL (/dashboard) as fallback

## Validation

- [x] Backend tests passed

Commands run:

```bash
make test
```

## Risk

- User-facing impact:
- Deployment or migration impact:
- Rollback approach:

## Notes

- Related issue:
- Follow-up work:
- Docs or operator runbooks updated when behavior changed:

### DIFF
--- a/server/skillhub-auth/src/main/java/com/iflytek/skillhub/auth/oauth/OAuth2LoginSuccessHandler.java
+++ b/server/skillhub-auth/src/main/java/com/iflytek/skillhub/auth/oauth/OAuth2LoginSuccessHandler.java
@@ -1,23 +1,29 @@
 package com.iflytek.skillhub.auth.oauth;
 
+import java.io.IOException;
+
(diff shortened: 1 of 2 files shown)


XiaoSeS:
补了回归测试 + 恢复 clearAuthenticationAttributes 对称性

PR #275, opened 2026-04-09
feat(ci): add PR batch test deployment workflow

wowo-zZ:
## Summary
- add a manual GitHub Actions workflow that batches selected PRs onto a shared Hong Kong manual-test runtime
- add helper scripts to synthesize the merge batch in CI and deploy it through a restricted remote wrapper
- document the required GitHub secrets, runtime metadata, and operator flow for the shared test environment

## What This Enables
- operators can choose a small set of non-conflicting PRs and deploy them together to `https://skill.xf-yun.com.cn`
- the shared runtime keeps a stable floating tag (`manual-test-hk`) plus an immutable tag for traceability
- the remote host records deployment metadata so testers can confirm exactly which batch is live

## Security Notes
- the HK test machine now uses a dedicated `skillhub-deploy` SSH user instead of direct `root` login from GitHub Actions
- `skillhub-deploy` has passwordless sudo only for `/usr/local/bin/skillhub-test-deploy`
- the deploy user cannot run arbitrary `sudo` commands and does not have direct Docker socket access
- repository secrets `TEST_RUNTIME_SSH_HOST`, `TEST_RUNTIME_SSH_USER`, and `TEST_RUNTIME_SSH_KEY` have already been configured

## Validation
- `bash -n scripts/prepare-pr-batch.sh`
- `bash -n scripts/deploy-test-runtime.sh`
- `bash -n scripts/skillhub-test-deploy-remote.sh`
- YAML parse of `.github/workflows/pr-batch-test-deploy.yml`
- verified `skillhub-deploy` can run the deployment wrapper but cannot run arbitrary `sudo` or direct `docker` commands
- performed a real rehearsal by tagging the current runtime images as `manual-test-hk` and deploying them through the restricted deploy user

### DIFF
--- a/.github/workflows/pr-batch-test-deploy.yml
+++ b/.github/workflows/pr-batch-test-deploy.yml
@@ -1,0 +1,161 @@
+name: PR Batch Test Deploy
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_numbers:
+        description: "Comma/newline separated PR numbers to merge onto the base branch"
+        required: true
+        type: string
+      base_ref:
+        description: "Base branch to build from"
+        required: false
+        default: main
+        type: string
+      deploy_channel:
+        description: "Floating image tag used by the shared HK test machine"
+        required: false
+        default: manual-test-hk
+        type: string
+
+concurrency:
+  group: pr-batch-test-runtime
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  packages: write
+  pull-requests: read
+
+env:
+  DOCKER_PLATFORM: linux/amd64
+
+jobs:
+  build-and-deploy:
+    name: Build And Deploy Manual Test Batch
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Ensure helper scripts are executable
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-batch-test-deploy.yml
```diff
@@ -0,0 +1,161 @@
...
+            --pr-list "${{ inputs.pr_numbers }}" \
+            --base-ref "${{ inputs.base_ref }}" \
+            --deploy-channel "${{ inputs.deploy_channel }}"
+
+      - name: Build and push backend image
+        uses: docker/build-push-action@v6
+        with:
+          context: ./server
+          file: ./server/Dockerfile
+          platforms: ${{ env.DOCKER_PLATFORM }}
```
dongjiang1989, line 81: suggest support muti-arch docker image

wowo-zZ: Thanks for the suggestion. For this workflow, the scope is limited to deploying the shared test environment, and the current target machine is `x86_64` only. Keeping the image build as `linux/amd64` makes the workflow faster and simpler for this use case. If we later need to support additional test machines with different architectures, we can extend this workflow to multi-arch at that point.

wowo-zZ:
FYI: the failing backend unit test is coming from the current `main` merge ref rather than the changes in this PR. This PR only adds the manual test deployment workflow and related scripts/docs. If the approach looks good, please feel free to review/approve based on the PR content. Thanks!

dongmucat (approved):

PR #324, opened 2026-04-17
Add k3s deployment mode — lightweight alternative to k8s for small/medium enterprises

riverfor:
## Summary

A new and simple deployment mode has been added: deployment under k3s, suitable for small and medium-sized enterprises.
Compared to k8s, k3s requires fewer resources.

## Validation
No code was changed; the following tests should be unnecessary. I successfully deployed and conducted simple tests on my local k3s cluster.

- [x] Backend tests passed

Commands run:

- [Install k3s](https://docs.k3s.io/installation)
- Install SkillHub on a k3s cluster
```bash
kubectl apply -k deploy/k3s/overlays/quickstart/
```

## Risk

None

## Notes

- Playbook: [deploy/k3s/README_zh.md](https://github.com/riverfor/skillhub/blob/main/deploy/k3s/README_zh.md)


Pipreola:
@riverfor Thanks for contributing the K3s deployment setup and for validating it on your local cluster. After careful review, we've decided to close this PR for the following reasons:

**1. K3s and the existing K8s deployment have no material difference at the manifest level**

K3s is a fully conformant Kubernetes distribution, and `deploy/k8s/overlays/with-infra/` already provides a one-shot deployment with bundled PostgreSQL and Redis that runs directly on a K3s cluster via `kubectl apply -k`. The only K3s-specific concern is switching the Ingress Controller from nginx to Traefik; everything else (StorageClass, resource limits, images) in the current k8s configuration is already K3s-compatible — for instance, none of the PVCs hardcode `storageClassName`, so they fall back to the cluster default (`local-path` on K3s) automatically.

**2. The copy-based overlay introduces long-term maintenance cost**

8 of the 11 YAML files in this PR are line-by-line copies of existing files under `deploy/k8s/`. Any future change to the k8s deployment would have to be mirrored here, or the two paths will drift. That drift has in fact already happened in this PR: `deploy/k3s/overlays/quickstart/backend-deployment.yaml` is missing the `OAUTH2_GITHUB_CLIENT_ID` and `OAUTH2_GITHUB_CLIENT_SECRET` environment variables that exist in `deploy/k8s/base/backend-deployment.yaml`, which would break GitHub OAuth login on K3s deployments.

Thanks again for the contribution and for your interest in the project — we appreciate it.

PR #413, opened 2026-05-09
fix(ci): pin pnpm 10.33 and approve esbuild build scripts

dongmucat:
## Background

The `publish-images` workflow's web image build started failing with:

```
#21 5.353 [ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: esbuild@0.25.12
#21 5.353 Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
ERROR: process "/bin/sh -c pnpm install --frozen-lockfile" did not complete successfully: exit code: 1
```

Example failed run: https://github.com/iflytek/skillhub/actions/runs/25588944735

## Root cause

Two compounding issues:

1. **pnpm version drift.** `web/package.json` did not pin a `packageManager`, so `corepack enable` resolved pnpm to the newest release on every build. The failing run logged ``Corepack is about to download .../pnpm-11.0.9.tgz``.
2. **pnpm 11 tightened postinstall approval.** Under `--frozen-lockfile`, unapproved build scripts (here `esbuild`) cause a non-zero exit instead of a warning. The approval list also needs to live in `pnpm-workspace.yaml`, which the Dockerfile never copied into the build stage.

## Fix

- `web/package.json`: pin `"packageManager": "pnpm@10.33.0"` so Corepack always provisions the same major version and the lockfile is read by a compatible client.
- `web/pnpm-workspace.yaml` (new): approve `esbuild` via `onlyBuiltDependencies` (pnpm 10 syntax, matching the pinned major).
- `web/Dockerfile`: include `pnpm-workspace.yaml` in the `COPY` before `pnpm install` so the approval list is visible at install time.

## Verification

Locally reproduced the failure with `corepack use pnpm@11.0.9` and the existing lockfile — same `ERR_PNPM_IGNORED_BUILDS: esbuild@0.25.12`.

After the fix, ran a full `docker buildx build --platform linux/amd64 --target build` on a clean cache. Install log showed:

- `Corepack is about to download .../pnpm-10.33.0.tgz`
- `.../esbuild@0.25.12/node_modules/esbuild postinstall: Done`
- `Done in 36.5s using pnpm v10.33.0`

`pnpm build` then completed and produced `dist/` as expected.

Also triggered the `publish-images` workflow against this branch to confirm end-to-end in CI: https://github.com/iflytek/skillhub/actions/runs/25589620667

## Impact

- CI-only change; no runtime or API-surface change.
- Frontend developers using a locally installed pnpm should have Corepack pick up `pnpm@10.33.0` automatically from `package.json`.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:22-alpine AS build
 RUN corepack enable
 WORKDIR /app
-COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
+COPY package.json pnpm-lock.yaml ./
 RUN pnpm install --frozen-lockfile
 COPY . .
 RUN pnpm build
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:22-alpine AS build
 RUN corepack enable
 WORKDIR /app
-COPY package.json pnpm-lock.yaml ./
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
 RUN pnpm install --frozen-lockfile
 COPY . .
 RUN pnpm build
--- a/web/package.json
+++ b/web/package.json
@@ -4,6 +4,11 @@
   "version": "0.1.0",
   "type": "module",
   "packageManager": "pnpm@10.33.0",
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "esbuild"
+    ]
+  },
   "scripts": {
     "install:ci": "pnpm install --frozen-lockfile",
     "dev": "vite",
--- a/web/package.json
+++ b/web/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.1.0",
   "type": "module",
+  "packageManager": "pnpm@10.33.0",
   "scripts": {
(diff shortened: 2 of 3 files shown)


dongmucat:
Thanks for the review @gemini-code-assist — applied in 04a45451:

- Moved `onlyBuiltDependencies: ['esbuild']` under the `pnpm` field in `web/package.json`
- Deleted `web/pnpm-workspace.yaml` (single-package project, no need to declare a workspace root)
- Reverted the `pnpm-workspace.yaml` addition to the Dockerfile `COPY` so it stays minimal

Verified locally with `docker buildx build --no-cache`: Corepack provisions pnpm 10.33.0, esbuild's postinstall executes (`esbuild postinstall: Done`), and `vite build` produces `dist/` as expected. End-to-end CI is green: https://github.com/iflytek/skillhub/actions/runs/25590088302

One caveat for future maintainers: pnpm 11 dropped support for the `pnpm` field in package.json and requires `pnpm-workspace.yaml` with `allowBuilds` instead. Since we're pinned to pnpm 10 via `packageManager`, this is fine today; whoever bumps to pnpm 11 will need to migrate the approval list back to a workspace file.

PR #534, opened 2026-06-16
fix(cli): treat highlighted install target as selected

SenLinLeo:
## Summary

  - Treat the currently highlighted install target as selected when the interactive multiselect prompt submits an empty selection.
  - Preserve the existing explicit multiselect behavior; only empty submissions use the highlighted fallback.
  - Add a unit test covering the Enter-on-highlighted-target path.

  ## Why

  When `skillhub install` detects multiple install targets, users can move the cursor to a target and press Enter expecting that row to be confirmed. In this flow, `prompts` can return an empty selection unless the user first checks an
  item, so the CLI currently treats it as cancellation.

  Falling back to the highlighted target makes Enter behave like confirm while keeping the existing multiselect flow intact.

## Tests

- `bun test test/unit/agents/resolver-interactive.test.ts test/unit/agents/resolver.test.ts`
- `bun run typecheck`
- `bun run lint`
- `bun run build`
- `bun test`

### DIFF
--- a/cli/src/agents/resolver.ts
+++ b/cli/src/agents/resolver.ts
@@ -160,6 +160,7 @@ function dedupeByRoot(candidates: AgentCandidate[]): AgentCandidate[] {
 
 async function selectTargetsInteractively(candidates: AgentCandidate[]): Promise<AgentCandidate[]> {
   const prompts = await import('prompts')
+  let highlightedIndex = 0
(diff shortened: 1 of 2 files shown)


SenLinLeo:
image
It has no impact on historical logic.

dongmucat (approved):
Thanks for the fix. LGTM.

PR #566, opened 2026-06-30
fix(frontend): patch undici alerts and harden staging web

dongmucat:
## What
- patch the 6 open Dependabot alerts on `web/pnpm-lock.yaml` by forcing `undici` to `>=7.28.0`
- refresh the web lockfile so `jsdom -> undici` resolves to `7.28.0`
- harden `make staging` so the web service uses a locally built `skillhub-web:staging` image instead of bind-mounted frontend files

## Why
GitHub currently reports 6 open Dependabot alerts on the default branch, all pointing to `web/pnpm-lock.yaml` and `undici@7.25.0` via `jsdom`.

While verifying the fix, `make staging` failed on this machine because the nginx container could list bind-mounted files but could not read them (`Operation not permitted`). Using the existing `web/Dockerfile` for the staging web service removes that environment-sensitive bind-mount path and keeps the staging gate reliable.

## How
- add `pnpm.overrides.undici@<7.28.0: ^7.28.0` in `web/package.json`
- regenerate `web/pnpm-lock.yaml` so the resolved `jsdom -> undici` edge moves from `7.25.0` to `7.28.0`
- keep the existing local `pnpm build` step in `make staging`, then also build `skillhub-web:staging`
- switch `docker-compose.staging.yml` web service from `nginx:alpine` + bind mounts to `skillhub-web:staging`

## Testing
- `cd web && npm_config_registry=https://registry.npmjs.org/ pnpm audit --audit-level=low`
- `cd web && pnpm install --frozen-lockfile && pnpm run typecheck && pnpm run lint && pnpm run test && pnpm run build`
- `make test-backend-app`
- `make staging`
- `git diff --check`

## Impact
- merging this PR should clear the current 6 open Dependabot alerts for `undici` on the default branch
- no API contract or business logic changes
- `make staging` now validates the web service through the same image packaging path used by `web/Dockerfile`, instead of relying on host bind mounts


dongmucat:
Superseded by #567 with the same patch carried on a DCO-signed commit so required checks can pass cleanly.

PR #588, opened 2026-07-20
feat(i18n): add Russian locale (web UI + MessageSource)

michael-xiii:
## Summary
- Add `web/src/i18n/locales/ru.json` (full UI string parity with `en.json`).
- Register `ru` in i18next and the language switcher («Русский»).
- Add `messages_ru.properties` for server MessageSource.
- Minor: Russian relative time in notifications; account-disabled detection for Russian messages.

## Notes
- CLI commands, package names, and brand strings intentionally left untranslated.
- Privacy/legal pages still resolve to `en`/`zh` only (follow-up).
- Glossary: skill → скилл, namespace → пространство имён, token → токен.
- Verified on a private POC stand (web build); key tree and `{{…}}` placeholders match `en.json`.

## Test plan
- [ ] Language switcher shows Русский; UI strings switch
- [ ] Navigator `ru` detects Russian
- [ ] Web locale parity tests (`ru-locale.test.ts`)
- [ ] Server with `messages_ru` returns Russian validation when `Accept-Language` is `ru`


FenjuFu (approved):
Nicely done — this is more rigorous than most locale PRs:

- `ru-locale.test.ts` enforces both **key-tree parity with English** and **interpolation placeholder preservation**, so the translation can't silently drift or drop a `{{var}}`.
- Both halves are covered: web `ru.json` and the server-side `messages_ru.properties`.
- The switcher and i18n config wiring are complete.

One behaviour change worth calling out explicitly so it's a conscious decision, not a drive-by: `language-switcher.tsx` changes the fallback from `'zh'` to `'en'`. That actually makes the switcher agree with the existing `fallbackLng: 'en'` in `i18n/config.ts` — previously the control could show "中文" as selected while i18next had already fallen back to English — so I read it as a consistency fix rather than a product change. Flagging it for the record; if anyone disagrees it belongs in a separate PR.

LGTM otherwise. CI was pending first-time-contributor approval; I've released it.

michael-xiii:
Thanks for the review and LGTM, @FenjuFu.

@dongmucat — could you share why this PR was closed without merge? The change was approved, CLA/DCO are signed, and the only red check was E2E failing at “Start full dev stack” (looks infra-related rather than locale-specific).

We’d really like to see this work land in the repository: Russian locale for the web UI (`ru.json` + switcher/i18n wiring) and server `messages_ru.properties`, plus the parity tests. Happy to reopen, rebase onto current `main`, and re-run CI if that’s helpful — just let us know what you’d like adjusted.

PR #601, opened 2026-07-23
fix(publish): delete review tasks of any status when replacing a version

### LINKED ISSUE #596: [Bug](publish): re-uploading a rejected version with the same version returns 500
## 问题描述

Skill 版本审核被驳回后，使用相同版本号重新上传会返回 HTTP 500，无法创建新的待审核版本。

## 复现步骤

1. 发布一个非 PRIVATE 的 Skill，版本号为 `1.0.0`。
2. 等待安全扫描完成，版本进入 `PENDING_REVIEW`。
3. 管理员驳回审核，使版本状态变为 `REJECTED`。
4. 使用相同 Skill 包、相同版本号 `1.0.0` 再次上传。

## 实际结果

第二次上传返回 HTTP 500：

```json
{
  "code": 500,
  "msg": "An unexpected error occurred"
}
```
原 REJECTED 版本仍保留，未创建新的版本记录。
## 期望结果
如果产品语义允许驳回版本复用版本号，则应删除/替换旧的可替换版本，并创建新的待审核版本。
若不允许复用，则应返回明确的业务错误，例如“请先删除被驳回版本或使用新的版本号”，而不是 HTTP 500。
## 根因分析
SkillPublishService.deleteReplaceableVersionArtifacts 在覆盖非 PUBLISHED 的旧版本时，只会删除状态为 PENDING 的 review_task：
reviewTaskRepository.findBySkillVersionIdAndStatus(version.getId(), ReviewTaskStatus.PENDING)
        .ifPresent(reviewTaskRepository::delete);
但审核驳回后，关联的 review_task 状态已变为 REJECTED，该记录仍通过外键引用旧 skill_version。随后删除旧版本时触发数据库外键约束异常，最终被包装为 HTTP 500。
## 修复建议
若保持“驳回版本可被同版本号重传覆盖”的语义，应在删除旧 skill_version 前清理该版本关联的全部 review task，而不是只删除 PENDING 状态的任务，例如使用：
reviewTaskRepository.deleteBySkillVersionIdIn(List.of(version.getId()));
同时补充回归测试：REJECTED 状态的 1.0.0 重新上传 1.0.0 后，应成功创建新的版本并进入扫描/待审核流程。

FenjuFu:
Fixes #596

## What

Allow a rejected skill version to be re-uploaded with the same version number without returning HTTP 500.

## Why

A rejected review task still references its skill version. The replacement path previously removed only a pending review task, so deleting the rejected version violated the review_task foreign key.

## How

- Delete every review task attached to a version that is actually replaceable.
- Restrict replacement to the lifecycle states DRAFT, SCAN_FAILED, UPLOADED, and REJECTED.
- Reject PUBLISHED, YANKED, SCANNING, and other non-replaceable states with error.skill.version.exists.
- Keep replacement cleanup and new review-task creation inside the normal publishFromEntries workflow.

## Testing

- SkillPublishServiceTest covers re-uploading a REJECTED version through the public publish entry point.
- A boundary test verifies that a YANKED version and its review history are not deleted.
- Playwright E2E publishes, waits for scanning, rejects, re-uploads 1.0.0, verifies a new pending review task, and verifies the old rejected task returns 404.
- Local SkillPublishServiceTest: 35 tests passed.
- Frontend typecheck, ESLint, and Vitest: passed.
- Real PostgreSQL E2E: pending GitHub CI.

## Impact

No API or schema changes. Published and yanked history remains immutable. The behavior change is limited to replacing versions in the existing deletable lifecycle states.


XiaoSeS (changes requested):
There are two blocking issues before this can be merged:

1. `deleteReplaceableVersionArtifacts` still rejects only `PUBLISHED`, while the new unconditional `deleteBySkillVersionIdIn(...)` also applies to `YANKED`. A yanked version is previously published history and may own an `APPROVED` review task; this change can now delete that audit history. It may then still fail with an FK violation if the version is referenced by `skill_tag.version_id` or `promotion_request.source_version_id`. Please restrict replacement to the lifecycle states that are actually replaceable (at minimum exclude `YANKED`) and add a boundary test.

2. The regression test does not exercise the reported behavior. It invokes a private helper through reflection and verifies Mockito calls, so it cannot prove that re-uploading a REJECTED version succeeds through the publish workflow or that the real `review_task` FK no longer produces HTTP 500. Please cover the public `publishFromEntries` replacement path (preferably with a database-backed integration test for the FK), asserting that the old rejected version/task are removed and a new pending-review version/task is created. This coverage should live with the existing `SkillPublishServiceTest` rather than a second reflection-based service test.

XiaoSeS (approved):
Re-reviewed the corrected head and all comments. Replacement is restricted to deletable lifecycle states, YANKED history is protected, the regression is covered through publishFromEntries, and the real PostgreSQL/Playwright flow verifies reject then same-version re-upload plus removal of the old review task. All required checks, DCO, and CLA pass.

XiaoSeS:
> *This was generated by AI during triage.*

Post-merge shared-runtime validation completed against deployed integration revision `b5607dfa`, which contains merge commit `4fdc7e3d`.

The real API/PostgreSQL/scanner/review flow passed:

1. A regular user published version `1.0.0` and reached `PENDING_REVIEW`.
2. A temporary test reviewer rejected the review task.
3. The same user re-uploaded the same skill/version successfully.
4. The replacement produced a different pending review-task ID while preserving the skill ID and version.
5. The old rejected review task returned HTTP 404.

Playwright 1.58.2 Chromium validation: 1/1 passed. Cleanup also passed: the temporary `SUPER_ADMIN` binding was removed, the generated skill was deleted, and backend health remained `UP`. No host or service restart was performed.

PR #619, opened 2026-07-29
[#618] Fix web ambient type isolation

### LINKED ISSUE #618: [Bug] Web typecheck is affected by unrelated ambient @types
### Summary

`web/tsconfig.json` does not declare `compilerOptions.types`, so TypeScript implicitly scans every visible package under `node_modules/@types` as an ambient type library.

That makes `typecheck` and `build` depend on workspace/cache cleanliness rather than only on `package.json` and the lockfile. A stale or partially restored `@types/*` directory can fail the build even though the package is not a project dependency.

Current configuration:

- [`web/tsconfig.json` has no explicit `types` list](https://github.com/iflytek/skillhub/blob/9f602f8184f23347124b3606ffce1a6ac5f19c0a/web/tsconfig.json#L1-L23)

### Steps To Reproduce

At `main` commit `9f602f8184f23347124b3606ffce1a6ac5f19c0a`:

```bash
cd web
pnpm install --frozen-lockfile
pnpm exec tsc --noEmit
mkdir -p node_modules/@types/skillhub_ghost_type
pnpm exec tsc --noEmit
```

The clean typecheck passes. With the unrelated stale directory present, TypeScript fails:

```text
error TS2688: Cannot find type definition file for 'skillhub_ghost_type'.
  The file is in the program because:
    Entry point for implicit type library 'skillhub_ghost_type'
```

This models shared CI workspaces or dependency caches that restore a partial/stale `@types` package.

### Expected Behavior

Typechecking should include only ambient type packages intentionally required by the web app.

Declare an explicit `compilerOptions.types` list (an empty list is sufficient if all required types are imported explicitly), and add a regression check that injects an unrelated ghost `@types` directory.

If Vite ambient types become necessary later, list them explicitly rather than re-enabling implicit discovery.

### Environment

- Branch: `main`
- Commit: `9f602f8184f23347124b3606ffce1a6ac5f19c0a`
- pnpm: `10.33.0`
- TypeScript resolved from the current lockfile

### API Contract Impact

None. Build/typecheck isolation only.

### Logs Or Screenshots

Reproduced locally with the commands above; clean typecheck passed and the injected ghost type failed with TS2688.

dongmucat:
## Summary
- declare an explicit empty `compilerOptions.types` list for the web project
- prevent unrelated or partially restored `node_modules/@types/*` directories from affecting typecheck and build results
- retain imported React, Vitest, and other dependency types without enabling Vite ambient globals

Closes ISSUE-92
Closes #618

## Validation
- injected `web/node_modules/@types/skillhub_ghost_type`, then `pnpm exec tsc --noEmit` — passed
- `make typecheck-web` — passed
- `make lint-web` — passed
- `make test-frontend` under Node 20.20.0 — 181 files / 618 tests passed
- `pnpm --dir web build` — passed

Playwright was not run because this compiler-only change does not touch `web/src` or UI behavior.

### DIFF
--- a/web/tsconfig.json
+++ b/web/tsconfig.json
@@ -9,6 +9,7 @@
     "allowImportingTsExtensions": true,
     "isolatedModules": true,
     "moduleDetection": "force",
+    "types": [],
     "noEmit": true,
     "jsx": "react-jsx",
     "strict": true,


FenjuFu (approved):
LGTM. Declaring an explicit `types: []` is the standard fix for ambient-`@types` bleed — it stops TS from implicitly pulling every `node_modules/@types/*` into the global scope, so typecheck/build depend on `package.json` + lockfile rather than workspace cleanliness. The types actually used (React, Vitest, etc.) come in through imports / `///` references, so they're unaffected — and CI's typecheck job is the direct proof this holds.

PR #624, opened 2026-07-29
fix(web): stop React 19 portal removeChild/insertBefore crashes

michael-xiii:
## Summary

Fixes recurring React 19 DOM crashes (`Failed to execute 'removeChild'/'insertBefore' on 'Node'`) that surface as TanStack Router’s default **Something went wrong! / Hide Error** overlay (or the i18n RouteError UI). The failure was not limited to one screen: after each spot fix it reappeared on another page that still used competing portals, route unmount, or browser DOM mutation under `#root`.

### Root cause

Two overlapping classes:

1. **Portal / overlay teardown** — hand-rolled `Dialog` (`createPortal` → `document.body`) and Radix **Select / DropdownMenu** portals (plus Sonner `flushSync` toasts) tore down nodes that React 19 no longer considered children of the same parent—especially when a Select/Dialog stayed open during `navigate()`, or list pages remounted cards while N× per-skill star queries resolved.
2. **Chrome / Google automatic translate** — built-in Translate rewrites text nodes under `#root` (console often shows `gstatic.com/_/translate_http/...` CSP blocks). Triggers large re-renders such as skill overview expand even in private windows without extensions. Worse when `<html lang="zh-CN">` does not match the in-app UI locale.

### Approach (shared UI canon, not page-by-page workarounds)

1. **Dialog** → `@radix-ui/react-dialog` adapter (same pattern as Select/DropdownMenu).
2. **Select / DropdownMenu Content** → render **in-tree (no Portal)** so Content unmounts with its trigger.
3. **Dialog / Sonner** stay on isolated `#skillhub-portals` host; overlays use `translate="no"`.
4. **LanguageSwitcher** → in-tree menu (same idea as UserMenu; no body portal).
5. **Dismiss** open overlays on **pathname** change via overlay/body pointer events (synthetic Escape is unreliable / `isTrusted === false`).
6. **`navigateAfterOverlays`** for leave paths that close Dialog then navigate (publish, skill delete, withdraw review).
7. **Search list**: one shared `['skills','stars']` highlight (`useStarredIdSet`) instead of N× `useStar`; deferred Sonner toasts; `keepPreviousData` while typing.
8. **RouteError** recoverable UI with Reset (defense against Hide-Error loops).
9. Layout: `overflow-x-clip` only on the decorative layer so in-tree Select is not clipped.
10. **Chrome Translate guard**: `translate="no"` / `notranslate` on `<html>`, `#root`, `#skillhub-portals`; `<meta name="google" content="notranslate">`; `syncDocumentLanguage(i18n)` so `lang` matches the UI locale; markdown overview `translate="no"`.
11. **Skill overview expand**: memoized `MarkdownRenderer` with stable `onLinkClick` (`useCallback`); quiet ResizeObserver via generation + double rAF; no `max-height` CSS transition.

### Out of scope / intentional exceptions

- Native `<select>` in create-token remains a local exception until that dialog is re-checked against this Dialog adapter.
- No server/scanner changes.

## Test plan

- [ ] `cd web && pnpm exec vitest run src/shared/ui/dialog.test.tsx src/shared/ui/select.test.ts src/shared/lib/dismiss-open-overlays.test.ts src/shared/lib/navigate-after-overlays.test.ts src/shared/lib/portal-container.test.ts src/shared/lib/document-language.test.ts src/features/skill/overview-collapse.test.ts src/features/skill/markdown-renderer.test.tsx`
- [ ] `cd web && pnpm exec tsc -b` / lint
- [ ] Manual `/dashboard/publish`: open namespace/visibility Select (not clipped), upload, precheck Confirm → success navigate without overlay
- [ ] Manual `/search` (logged in): type query, scroll results; no removeChild
- [ ] Manual skill detail: **expand overview**; report/delete/withdraw Confirm → navigate
- [ ] Manual `/dashboard/tokens`: create-token dialog
- [ ] Manual header LanguageSwitcher / UserMenu; ConfirmDialog ESC / overlay / X
- [ ] Console: no `translate_http` / gstatic translate stylesheet attempts after load + overview expand
- [ ] Optional: Incognito if a residual crash remains (extensions can still mutate DOM)

## Commits

- `fix(web): replace custom Dialog with Radix to stop portal races`
- `fix(web): stop portal removeChild races on search and shell`
- `fix(web): systemic overlay teardown for React 19 removeChild`
- `fix(web): harden overlay dismiss after code review`
- `fix(web): batch starred highlight and defer Sonner toasts`
- `fix(web): block Chrome Translate DOM races on skill overview`

Signed-off-by: Mikhail Neradkov <michael.neradkov@gmail.com>


FenjuFu (approved):
Reviewed the approach and the shape of the change. LGTM on methodology — approving, with one honest caveat below.

**Why this is the right framing:** the recurring `removeChild`/`insertBefore` crashes under React 19 aren't a one-screen bug; they come from portals/overlays (hand-rolled `Dialog`, Radix Select/DropdownMenu, Sonner `flushSync`) tearing down nodes React 19 no longer treats as children of the same parent, plus browser DOM mutation under `#root`. Chasing them page-by-page is why earlier spot fixes kept resurfacing elsewhere. Centralizing on `portal-container` / `dismiss-open-overlays` / `navigate-after-overlays` and routing all overlays through it is the correct systemic fix rather than another spot patch.

**What gives me confidence:** the key CI jobs are green (Web Build & Test, Server Unit, Docs), and the PR ships focused unit tests for each new primitive (`portal-container`, `dismiss-open-overlays`, `navigate-after-overlays`, `document-language`, `overview-collapse`, dialog/select) rather than just asserting the crash is gone.

**Honest caveat for the maintainers, not a block:** this is 37 files touching every overlay/portal/route-unmount path in the web app. The blast radius is real — I'd want at least one manual pass through the actual app (open/close dialogs, dropdowns, toasts, then navigate mid-overlay on a few routes) before merge, since that interaction surface is exactly what unit tests approximate but don't fully reproduce. The design is sound; the ask is a smoke-test on the live UI given the breadth.

Issue #628, opened 2026-07-30
feat(auth): build a unified identity federation architecture


XiaoSeS:
## What to build

Introduce a staged identity federation architecture that keeps protocol verification in provider adapters while centralizing trusted-provider resolution, authority pinning, subject resolution, access/provisioning policy, identity binding, account-state guards, profile synchronization, principal construction, audit, and session creation in SkillHub.

The formal design is tracked in `docs/21-unified-identity-federation-design.md` and will be reviewed separately before the core migration is merged.

## Acceptance criteria

- [ ] Existing GitHub, GitLab, OIDC, local password, Device Flow, and API token behavior remains compatible.
- [ ] Every external provider returns protocol facts only and cannot directly create accounts, roles, principals, or sessions.
- [ ] Provider code is pinned to a normalized authority and mismatches fail closed.
- [ ] External subjects are typed; aliases, conflicts, and concurrent first login have explicit rules.
- [ ] Email is not used for implicit account takeover or merge.
- [ ] Profile sync, identity link, and account merge are independent, auditable policies.
- [ ] LDAP/AD, DingTalk, and CAS adapters are implemented only after the shared adapter contract is frozen.
- [ ] Each implementation PR is integrated into `big-main` and validated on the test environment before maintainers consider merging to `main`.

## Delivery order

- P0: close current identity safety gaps.
- P1: unified identity core, Binding V2, profile/provisioning policy, provider registry, identity link, and safe account merge.
- P2: LDAP/AD, DingTalk, and CAS adapters.
- P3: broker guidance and trusted gateway assertions.
- P4: SCIM lifecycle provisioning and explicit entitlement mapping.

Existing requirements remain the source for their protocol slices: #613, #583, #456, and #260. Community PRs #437 and #467 stay open while the shared contract is stabilized.

## Blocked by

None - can start immediately.

XiaoSeS:
Stage 0A PR is open: #688. This one only covers trusted OAuth identity attribute hardening; 0B and 0C will be separate follow-up PRs.

Issue #583, opened 2026-07-16
Add DingTalk OAuth login provider


gtsz666:
## Motivation

Some enterprise SkillHub deployments use DingTalk as their primary identity provider. Adding an optional DingTalk OAuth login provider would make self-hosted deployments easier to adopt in DingTalk-centered organizations while preserving the existing GitHub/GitLab OAuth, local account, mock auth, API token, and session flows.

## Proposed scope

- Add a `dingtalk` OAuth provider that is disabled unless explicitly configured.
- Reuse the existing Spring Security OAuth2 login flow, including state validation, success/failure handlers, session establishment, RBAC principal construction, and audit/session behavior.
- Reuse the existing identity mapping model: `identity_binding(provider_code, subject)` remains the unique external identity key.
- Map DingTalk user identity using the most stable returned identifier available, preferring `unionId`, then `openId`, then `userId` when applicable.
- Do not assume DingTalk always returns email. Treat email/mobile as optional profile fields, not primary identity keys.
- Surface the provider through the existing auth method/provider discovery so the frontend can render the login entry dynamically.
- Document runtime configuration for local, Docker/Compose, and Kubernetes deployments.

## Out of scope

- No company-specific intranet configuration, tenant-specific organization rules, or real credentials in the repository.
- No DingTalk contact synchronization, department-to-namespace automation, approval workflow, robot notification, mobile-only H5 login, or internal deployment scripts.
- No replacement of existing authentication mechanisms.

## Implementation direction

Current code already has the main extension points needed:

- `CustomOAuth2UserService` delegates OAuth login to `OAuthLoginFlowService`.
- `OAuthClaimsExtractor` handles provider-specific profile mapping.
- `IdentityBindingService` binds or creates platform users from normalized OAuth claims.
- `AccountMergeService` can consolidate identity bindings, API tokens, roles, namespace memberships, and local credentials.
- `/api/v1/auth/methods` and `/api/v1/auth/providers` already advertise configured OAuth methods to the frontend.

The minimal contribution would add DingTalk-specific provider configuration and claims extraction/token-userinfo adaptation while keeping provider-specific logic behind the existing auth extension points.

## Rollout and security

- Default behavior remains unchanged unless the DingTalk client registration is configured.
- `client_id`, `client_secret`, and tenant/app-specific values must be injected by environment variables or Kubernetes Secrets.
- CI tests should mock DingTalk endpoints and must not depend on real DingTalk network credentials.
[...]

Issue #260, opened 2026-04-08
[Feature] 能否支持企业内部的ldap登录


cw1427:
### Problem

[Feature] 能否支持企业内部的ldap登录

### Proposed Solution

[Feature] 支持企业内部的ldap登录 可通过环境变量或者配置文件配置ldap 地址

wowo-zZ:
## Issue Triage

- Route: `triage/deferred`
- Priority: `priority/p2` (2.6/5)
- Effort: `effort/l` (4/5)
- Confidence: `3/5`
- Risk: `low`
- Analysis Mode: `rules-only`

### Why
- Open for 8 days; entering the 2-week closure warm-up window.
- Thread already has follow-up discussion, so backlog pressure is rising.

### Next Action
- Keep the issue in the deferred queue and let the 6-hour rescore keep lifting it; it is forced into an active lane by day 10. If it is still open on day 14, treat it as a P0 escalation target under the SLA and prioritize it in the next triage pass.

---

## 问题分流结果

- 路由: `triage/deferred`
- 优先级: `priority/p2` (2.6/5)
- 修复投入: `effort/l` (4/5)
- 信息完整度: `3/5`
- 风险: `低`
- 分析模式: `rules-only`

### 原因
- 已打开 8 天，开始进入 2 周闭环预热窗口
- 已有后续讨论，积压压力在上升

### 下一步
- 将 issue 保留在 deferred 队列，并由 6 小时一次的 rescore 持续抬升；最晚在第 10 天强制进入 active lane。若第 14 天仍未闭环，应按 SLA 视为 P0 升级目标，并在下一次 triage 中重点处理

cw1427:
已经手动用AI加上了LDAP的支持 晚点push PR上去

PR #630, opened 2026-07-30
docs(auth): define unified identity federation architecture

XiaoSeS:
## What

- add the formal unified identity federation architecture
- define trusted provider resolution, authority locking, typed subjects and aliases, policy boundaries, identity linking, and safe account merging
- define native-versus-broker protocol support for OAuth/OIDC, LDAP/AD, DingTalk, CAS, SAML, trusted gateways, Kerberos, and SCIM
- mark the current Direct/Passive principal-returning extension points as compatibility-only
- define phased PR, migration, rollback, test, `big-main`, and test-environment gates

## Why

SkillHub has multiple protocol requests (#260, #456, #583) and community implementations (#437, #467), but the current OAuth-centered claims and Direct/Passive SPIs do not provide one enforceable identity safety boundary.

This PR provides the architecture and delivery contract for #628 before protocol-specific implementations define incompatible identity models.

## How

The design is based on established patterns from Keycloak, authentik, Dex, Authelia, Backstage, Grafana, Kubernetes, Spring Security, and the relevant OIDC, OAuth, LDAP, CAS, SAML, SCIM, and WebAuthn standards. It intentionally uses a small core with capability-specific adapters instead of a universal provider interface.

## Testing

- `git diff origin/main...HEAD --check`
- verified all four documentation files are isolated on a clean branch
- documentation-only change; no runtime behavior changed

## Impact

No API, database, deployment, or runtime behavior changes. Implementation remains split into independently reviewable follow-up issues and PRs.


FenjuFu:
Read through the design doc and skimmed the four implementation PRs it anchors (#631, #633, #639, plus #644 which is separate). First: this is an unusually thorough piece of work — surveying Keycloak / authentik / Dex / Authelia / Spring Security, laying out the A/B/C options and landing on the phased hybrid, and naming the concrete gaps (typed subjects, authority pinning to issuer/directory/entityID, cross-issuer impersonation, safe merge) is exactly the right level of rigor for something that touches the auth boundary.

I want to be straight about my role here, though: **I'm not going to single-approve this group.** #631 (trusted external identity attributes), #633 (global membership on approval), and #639 (isolating the unsafe merge flow) are all authentication-boundary changes where a review mistake is a security hole, and the direction itself — a unified federation model marked as superseding the current Direct/Passive extension points — is a product/architecture decision that a maintainer or security owner should sign off on, not a passing reviewer. So treat my notes as input, not a gate.

Two things I did want to raise from the read:

1. **#631 email assurance:** requiring a *verified* entry from `/user/emails` before deriving GitHub email assurance, and refusing to let unverified external email populate/overwrite the platform email, is the correct hardening — this is a classic account-takeover vector (unverified provider email → email-domain access). Worth confirming the same "must be verified" rule is enforced for every provider the federation model onboards (OIDC `email_verified`, DingTalk, SAML assertions), not just GitHub, so the guarantee is uniform rather than per-provider.

2. **#639 fail-closed:** returning a single localized `503` on the three legacy merge routes and deleting `AccountMergeService` is the safe move. The one thing to double-check is that nothing else in the codebase still calls into the removed destructive migration path — a compile-time removal is good, but a reflective/config-driven caller wouldn't show up as a build break.

Sequencing (phased PR + migration + rollback + `big-main` gates) is laid out well. Deferring to the maintainers on whether to adopt this federation direction and on the security review of the implementation PRs.

PR #635, opened 2026-07-30
feat(builtin-skills): add reviewed starter collection

wowo-zZ:
## Summary

- add 15 reviewed, general-purpose built-in Skill sources with pinned upstream commits, licenses, attribution, and safety adaptations
- add one realistic regression case per Skill plus deterministic ZIP packaging and provenance validation
- enforce the collection in CI and through SkillHub's production package validator

## Validation

- [x] `make test-builtin-skills`
- [x] `BACKEND_TEST_JAVA_OPTIONS='-XX:+EnableDynamicAgentLoading -Duser.language=en -Duser.country=US' make test-backend-app` — 658 tests, 0 failures, 1 skipped
- [x] Cisco AI Skill Scanner 1.0.2 static scan — 15/15 SAFE, 0 findings
- [x] real FFmpeg extraction plus invalid-index, conflicting-option, and no-clobber checks
- [x] `make lint-web`
- [ ] `make typecheck-web` — local dependency refresh was blocked by repeated npm registry ECONNRESET; no frontend files changed
- [x] OpenAPI regeneration not required; no API contract changed

## Risk

- no runtime manifest, database, API, or deployment behavior changes in this PR
- generated ZIPs stay ignored; source, licenses, notices, catalog, and evals are reviewable in-repo
- rollback: revert this commit

## Follow-up

Stacked draft #636 adds SHA-256 verification, release configuration, and the deployment toggle. It will receive the 15 runtime manifest entries after immutable CDN URLs are available.

PR #636, opened 2026-07-30
feat(bootstrap): verify built-in skill artifacts

wowo-zZ:
## Summary

- require a lowercase SHA-256 for every built-in Skill manifest item and verify downloaded ZIP bytes before extraction
- publish the 15 reviewed starter packages through immutable official CDN URLs with exact SHA-256 values
- expose `SKILLHUB_BUILTIN_SKILLS_ENABLED` in release Compose and env templates, defaulting to `true`
- validate that the runtime manifest contains all 17 built-in Skills and matches deterministic build artifacts
- cover version upgrades where existing system Skills are idempotent and a user-owned `@global` slug conflicts

## Validation

- [x] `make test-builtin-skills`
- [x] `./scripts/tests/validate-release-config-test.sh`
- [x] Java 21 backend regression — 664 tests plus 9 isolated auth-filter tests, 0 failures, 1 skipped
- [x] upgrade conflict integration — `total=17, published=14, idempotentSkipped=2, conflictSkipped=1, failed=0`
- [x] conflict regression preserves the user-owned `weather` Skill and publishes the other 14 new Skills through the real publish/JPA path
- [x] `make typecheck-web` and `make lint-web`
- [x] frontend unit regression — 181 files, 618 tests, 0 failures
- [x] all 15 public CDN artifacts downloaded and matched their pinned SHA-256 values
- [x] clean release deployment from `main-0731@2a7b110af58f2bff06f564c1aaa0f2b2ae755fcc` — built-in sync `total=17, published=17, failed=0`
- [x] rolling upgrade to `main-0731@5da29f45a58b3b47a8b4a64126bea3653593cbf9` with existing volumes — `published=0, idempotentSkipped=17, conflictSkipped=0, failed=0`
- [x] all 17 slugs remain publicly discoverable; backend, web, scanner, and password login are healthy
- [x] cloud image build succeeded: https://github.com/iflytek/skillhub/actions/runs/30626078031
- [x] server, web, and scanner OCI revisions all equal `5da29f45a58b3b47a8b4a64126bea3653593cbf9`

## Risk

- existing custom manifest entries now need a valid `sha256`; the bundled entries are migrated here
- release deployments initialize built-in Skills by default and can opt out with `SKILLHUB_BUILTIN_SKILLS_ENABLED=false`
- on upgrade, an identical system-owned version is skipped; a user-owned `@global` slug is preserved and reported as a conflict while remaining manifest items continue
- the 15 new artifact paths include version and SHA-256; later changes must publish a new immutable artifact
- rollback: revert this PR and disable built-in synchronization if an operator needs to prevent future initialization

## Dependency

Stacked on #635. The immutable CDN upload, runtime manifest entries, checksum verification, cloud image build, clean-deployment validation, rolling-upgrade validation, and upgrade-conflict regression are complete.


FenjuFu (approved):
LGTM. This is the right supply-chain hardening for CDN-hosted built-in skills, and the verification chain is airtight:

- The manifest loader enforces the checksum format up front — `Pattern.compile("[0-9a-f]{64}")` requires lowercase 64-hex, rejecting the item otherwise.
- At install time the exact downloaded ZIP bytes are hashed and compared against the pinned value **before** extraction or publication, so a changed or overwritten CDN object is skipped rather than trusted.
- Because the loader forces lowercase and `sha256(bytes)` emits lowercase, the `equals` comparison has no case-mismatch blind spot.

Nice side benefit: this also plumbs `SKILLHUB_BUILTIN_SKILLS_ENABLED` through `compose.release.yml` and the env templates — that's the exact gap I ran into on #593 (the flag was read by `application.yml` but never passed through release Compose, so setting it did nothing). Good to see it fixed at the source here. Tests cover the loader and initializer paths.

PR #688, opened 2026-08-05
fix(auth): enforce trusted OAuth identity attributes

XiaoSeS:
What
- Harden external OAuth identity handling so only trusted, verified attributes can drive login and binding.
- Reject merged and system accounts on external login.
- Preserve authoritative email handling in identity binding.

Why
- Stage 0A of #628. This removes the current security gap before the later 0B/0C phases.

How
- Tighten GitHub claims extraction and identity binding checks.
- Add explicit exceptions for merged/system account login attempts.
- Extend auth tests to cover the new guardrails.

Testing
- `cd server && ./mvnw -pl skillhub-auth -am -DskipTests install`
- `cd server/skillhub-auth && ../mvnw test -Dtest=IdentityBindingServiceTest,OAuthLoginFlowServiceTest,GitHubClaimsExtractorTest,AccessPolicyTest`

Impact
- No API contract change.
- This is the first staged implementation PR; 0B and 0C will follow as separate PRs.

### DIFF
--- a/docs/02-domain-model.md
+++ b/docs/02-domain-model.md
@@ -245,6 +245,7 @@
 | avatar_url | varchar(512) | |
 | status | enum | `ACTIVE` / `PENDING` / `DISABLED` / `MERGED` |
 | merged_to_user_id | varchar(128) | 合并目标用户 ID，仅 MERGED 状态有值 |
+| system_account | boolean | 系统服务账号，禁止交互式 Web/OAuth 登录 |
 | created_at | datetime | |
 | updated_at | datetime | |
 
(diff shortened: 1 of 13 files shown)


FenjuFu:
Review finding (blocking): `GitHubClaimsExtractor` still uses `@Component`, but this patch removes the `org.springframework.stereotype.Component` import. The class will not compile unless the import is restored (or the annotation is intentionally removed and bean registration is handled another way). Please fix this and rerun the server tests. The CLA check is also still pending.

FenjuFu (commented):
Rechecked current head 833270b. The Component import is restored, so the earlier compilation blocker is resolved. The identity-binding path now rejects system, merged, pending, and disabled accounts before profile mutation; only verified email is persisted; and the focused tests plus Server, E2E, Web, Docs, Dependency Review, and DCO checks pass. I found no new code blocker. This is not an approval yet because CLA remains pending and the branch is currently unmergeable; recheck the final head after those gates are resolved.

XiaoSeS:
Maintainer review — changes required before merge:

1. GitHub `/user/emails` is now the trust source, which is correct for verified-email assurance. Please make the API base URL configurable for GitHub Enterprise or an internal proxy; the current hard-coded `https://api.github.com` is not deployable in all domestic networks.
2. Define the failure policy for 403/scope-missing/timeout/API outage. A failed lookup must remain `emailVerified=false` and must never fall back to trusting the OAuth profile email.
3. Add focused tests for those failure modes and document the required `user:email` scope.

The subject-based identity binding can remain usable; only email-dependent policy/binding decisions should fail closed.

PR #689, opened 2026-08-05
fix(smoke): support separate actuator target

### LINKED ISSUE #627: test(smoke): support deployments where actuator endpoints are not exposed by the public ingress
> *This was generated by AI during triage.*

## Problem

`scripts/smoke-test.sh` uses one `BASE_URL` for both public application APIs and internal actuator checks. This produces a misleading failure when a deployment exposes `/api/*` through the public web ingress but keeps actuator endpoints on the backend service.

Observed in the test environment:

- Direct backend `http://127.0.0.1:8080/actuator/prometheus` returns `401`, matching `RouteSecurityPolicyRegistry`, `PrometheusSecurityTest`, and the smoke expectation.
- Public `https://skill.xf-yun.com.cn/actuator/prometheus` returns `200 text/html`, because `web/nginx.conf.template` does not proxy `/actuator/*` and the SPA fallback serves `index.html`.
- The same public fallback also returns HTML for `/actuator/health`.

This is not evidence that Prometheus metrics are anonymously exposed, but the current smoke output makes it look like a security regression.

## Expected behavior

Smoke testing should model the public application URL and the internal actuator URL separately, and must not treat an SPA fallback as an actuator response.

## Suggested direction

- Add a separate actuator/backend base URL (or an explicit actuator-skip mode) while preserving the 401 assertion against the real backend endpoint.
- Validate response content type/body for health and Prometheus checks so `200 text/html` cannot be accepted as actuator health.
- Document the intended command for Compose, staging, and ingress deployments.
- Keep `/actuator/prometheus` protected for anonymous backend requests.

## Acceptance criteria

- Public API smoke checks can run against a web ingress that does not expose actuator routes.
- Actuator checks can target the backend service independently.
- Anonymous access to the actual Prometheus endpoint is still asserted as `401`.
- SPA fallback responses are reported as routing/target errors, not successful actuator responses.
- Script-level tests cover split public/backend URLs.

ShinyHero666:
## Summary

- Split the public application target from the Actuator target via `ACTUATOR_BASE_URL`, while preserving the current single-URL default.
- Validate that health responses are Actuator JSON and reject HTML SPA fallbacks as routing/target errors.
- Keep the Prometheus anonymous-access assertion on the actual Actuator target and reject HTML responses.
- Extend script-level regression coverage for split targets and the HTML fallback failure case.
- Document local/Compose, staging, and ingress usage.

## Validation

- [ ] Backend tests passed (not applicable: shell/docs-only change)
- [ ] Frontend typecheck/build passed (not applicable: no frontend source changes)
- [x] OpenAPI SDK regenerated or checked when API contracts changed (no API contract changes)

Commands run:

```bash
bash -n scripts/smoke-test.sh
bash -n scripts/tests/smoke-test-admin-mode-test.sh
bash scripts/tests/smoke-test-admin-mode-test.sh
git diff --check
```

## Risk

- User-facing impact: smoke tests can now target private Actuator endpoints independently and no longer accept an HTML SPA fallback as health.
- Deployment or migration impact: none; `ACTUATOR_BASE_URL` defaults to the existing public/API target.
- Rollback approach: revert this commit to restore single-target status-only checks.

## Notes

- Related issue: Fixes #627
- Follow-up work: none
- Docs or operator runbooks updated when behavior changed: README smoke-test instructions updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,18 @@ Run it against a local backend:
 ./scripts/smoke-test.sh http://localhost:8080
 ```
 
+Local Compose and staging runs can keep using one backend URL. For an ingress
+deployment where the public URL exposes application APIs but keeps Actuator on
+the backend service, set a separate Actuator target:
+
+```bash
+ACTUATOR_BASE_URL=http://skillhub-server:8080 \
+  ./scripts/smoke-test.sh https://skillhub.example.com
+```
+
+The health check requires an Actuator JSON response, so an HTML SPA fallback is
+reported as a routing or target error instead of a successful health response.
+
 Admin label-management smoke checks run only when current admin credentials are
 supplied explicitly:
 
--- a/scripts/smoke-test.sh
+++ b/scripts/smoke-test.sh
@@ -2,16 +2,18 @@
 set -euo pipefail
 
 BASE_URL="${1:-http://localhost:8080}"
+ACTUATOR_BASE_URL="${ACTUATOR_BASE_URL:-$BASE_URL}"
 PASS=0
 FAIL=0
-COOKIE_JAR="$(mktemp)"
+TMP_DIR="$(mktemp -d)"
+COOKIE_JAR="$TMP_DIR/cookies"
 USERNAME="smoketest_$(date +%s)"
 EMAIL="${USERNAME}@example.com"
 PASSWORD="Smoke@2026"
 NEW_PASSWORD="Smoke@2027"
 
 cleanup() {
-  rm -f "$COOKIE_JAR"
+  rm -rf "$TMP_DIR"
 }
 
 trap cleanup EXIT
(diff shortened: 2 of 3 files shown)


FenjuFu (approved):
Reviewed the split-target behavior and regression coverage. ACTUATOR_BASE_URL preserves the existing default, both Actuator probes use the private target when configured, HTML SPA fallbacks are rejected, and the application API checks remain on BASE_URL. The focused shell tests plus full CI, DCO, and CLA are green. No blocking issues found.

FenjuFu (approved):
Verified against #627. The PR separates public API and Actuator targets, retains the anonymous Prometheus 401 assertion, rejects HTML SPA fallbacks, documents deployment usage, and adds split-target regression coverage. The script regression test passes locally.

PR #712, opened 2026-08-13
feat(namespace): auto-provision a personal namespace on registration

### LINKED ISSUE #711: [Feature] Optionally give each new account its own namespace on registration
### Problem

On a self-hosted instance, a brand-new account has nowhere of its own to publish.
It can only push into `global`, or wait for an administrator to create a namespace
and add it as a member. For an internal registry where everyone is expected to
publish, that is friction on every single onboarding.

### Proposed Solution

An operator-controlled policy: when a new account becomes usable, give it a
namespace it owns. The account is the only member and holds `OWNER`. The
namespace slug and display name come from templates the administrator can edit,
e.g. `${username}` or `${username}-space`.

Off by default, so upgrading changes nothing for existing deployments.

Note on wording: namespaces have no visibility flag today — only `GLOBAL` and
`TEAM` — and skill visibility is a property of each skill. So "personal" here
means a team namespace whose only member is that account, not a new privacy
concept. Namespace-level visibility would be a separate feature.

### Alternatives Considered

- **Documentation only** — tell operators to create namespaces by hand. Does not
  scale past a handful of users, and the admin namespace management added in #707
  still needs one action per person.
- **Always create one** — no toggle. Rejected: it would change behaviour for every
  existing deployment on upgrade.
- **Config-file only** — no runtime control. Workable, but requires a redeploy to
  change, and the naming templates are exactly the kind of thing an operator wants
  to try and adjust.

### Impact

This needs somewhere to store a setting an operator can change at runtime, and
SkillHub currently has no such mechanism: per-deployment knobs live in
`application.yml`, and the only stored preferences are per-user notification
preferences.

So the change comes in two parts: a small generic `system_setting` store, and the
feature that consumes it. The store is deliberately reusable — the most obvious
next consumer is #318 (an administrator switch for local registration), which
today can only be done by blocking `/api/v1/auth/local/register` at the gateway.

### Contract Or SDK Impact

Two new admin endpoints under `/api/v1/admin/settings/personal-namespace`
(`SUPER_ADMIN`). No changes to existing contracts, and no CLI impact.

---

I have an implementation ready and will open a PR against this issue.

Vast-Stars:
Closes #711

## Motivation

On a self-hosted instance a brand-new account has nowhere of its own to publish —
only `global`, or a namespace an administrator creates for it by hand. This adds
an operator-controlled policy that gives each newly activated account a namespace
it owns.

Doing that needs a setting an operator can change at runtime, and SkillHub has no
mechanism for one. So this comes as two commits that can be reviewed
independently:

1. **`feat(admin): add system settings storage`** — a generic store, no consumer.
2. **`feat(namespace): auto-provision a personal namespace on registration`** —
   the first consumer, plus the admin console surface.

Happy to split these into two PRs if you would rather review them separately; I
kept them together because a settings store with no caller is hard to judge.

## Scope

**Off by default.** `skillhub.namespace.personal-provisioning.enabled` defaults to
`false`, so upgrading changes no behaviour.

**"Personal" means what the model supports.** Namespaces have no visibility flag —
only `GLOBAL` and `TEAM` — and skill visibility is a property of each skill. So the
namespace created here is an ordinary team namespace whose only member is that
account, holding `OWNER`. Namespace-level visibility would be a separate feature.

### Why an `AFTER_COMMIT` listener rather than a call next to `GlobalNamespaceMembershipService.ensureMember`

That would have been the obvious spot, since it is already called from both
account-creation paths. But `namespace.created_by` and `namespace_member.user_id`
both reference `user_account(id)`, which rules out each variant:

- **Joining the registration transaction** lets a slug clash roll the registration
  back, so a namespace failure costs the user their account — or, on OAuth, their
  login.
- **Suspending it with `REQUIRES_NEW`** leaves the new transaction unable to see
  the uncommitted `user_account` row, so the foreign key check blocks on the outer
  transaction's row lock and the two wait on each other.

Provisioning after the account is committed avoids both. The listener is
deliberately **not** `@Async`, so the namespace exists by the time the user's next
request arrives, and it swallows failures — a naming clash or a database hiccup
should cost a namespace, not a login.

### Three trigger points

`UserActivatedEvent` is published wherever an account first becomes usable:

| Path | Location |
|------|----------|
| Local registration | `LocalAuthService.register` |
| External identity, first login | `IdentityBindingService.bindOrCreate` (ACTIVE only) |
| Administrator approval / re-enable | `AdminUserAppService.updateUserStatus` (transition into ACTIVE) |

The third matters for deployments gated behind approval: those accounts are
created `PENDING` at the first OAuth attempt and only become usable when an
administrator approves them. Re-enabling a disabled account publishes the event
again, so provisioning is idempotent — owning any non-global namespace skips it.

### Naming

Templates over `${username}`, `${email_prefix}` and `${user_id}`. Unknown
placeholders are left in place so a typo is visible rather than silently dropped;
`${username}` falls back to the email local part and then to the user id.

Slugs go through the existing rules, which is why the console renders a live
preview: underscores are not legal in a slug, so `${username}_space` yields
`alice-space`. A taken or reserved candidate gets a numeric suffix, so a user
named `admin` gets `admin-2`.

`SlugValidator` gains two non-throwing helpers (`normalize`, `isValid`) for
generated candidates; `slugify` keeps its exact previous behaviour.

### Why the templates are not in `application.yml`

They contain `${...}`, which Spring would resolve as property references, and
Boot 3.2 predates placeholder escaping. Only the enable flag lives in YAML;
template defaults come from `PersonalNamespaceProvisioningProperties` and are
edited in the console.

## Settings storage

One row per setting group, value as JSONB, so a group gains fields without a
migration. Reads take the caller's defaults:

```java
<T> T get(String settingKey, Class<T> type, T defaults)
```

That gives two properties worth keeping:

- A group nobody has overridden has **no row**, and resolves to whatever the
  deployment configured. Configuration-file-only deployments keep working exactly
  as before.
- A stored document that can no longer be parsed **also falls back** to the
  defaults, with a warning. One malformed row must not take down login.

Groups ignore unknown fields so a rolling upgrade can read documents written by a
newer node.

## Rollout impact

- New migration `V44__system_setting.sql`; additive, no changes to existing tables.
- Two new `SUPER_ADMIN` endpoints under `/api/v1/admin/settings/personal-namespace`;
  no existing contract changes, no CLI impact.
- Updating the policy writes an audit entry with the before and after.
- Default-off, so an upgrade with no operator action is a no-op.

## Verification

- `make test` — backend `BUILD SUCCESS`; frontend 193 files / 679 tests
- Backend boots against a real PostgreSQL and `V44__system_setting.sql` applies
- `scripts/check-openapi-generated.sh` used to regenerate
  `web/src/api/generated/schema.d.ts`, which is committed
- `pnpm typecheck` and `pnpm build` clean

## Follow-up this unblocks

`system_setting` is generic. The most direct next consumer is #318 — an
administrator switch for local registration, which today can only be done by
blocking `/api/v1/auth/local/register` at the gateway. I am happy to send that as
a follow-up if you want it.


Vast-Stars:
Pushed another commit: `feat(namespace): let operators choose which namespaces new accounts join`.

Same story as the backfill commit — found by running this on our own intranet
registry, and it will hit anyone who does the same.

**A namespace you create to replace `global` is invisible to everyone.**
`listNamespaces` only returns namespaces the caller is a member of, and the
only thing that ever added members automatically was hard-wired to the slug
`global`. So an operator stands up an organisation-wide namespace, and nobody
can see it — including, confusingly, in the "my namespaces" list they would
use to check.

This makes that list a setting:

- `namespace.default-membership` holds the slugs every newly activated
  account is enrolled in. Defaults to `["global"]`, which is what every
  deployment did before, so upgrading changes nothing.
- `GlobalNamespaceMembershipService` becomes
  `DefaultNamespaceMembershipService`, since it no longer means one specific
  namespace.
- Saving validates each slug resolves to an ACTIVE namespace, so a typo fails
  at the moment it is made. Enrolling tolerates a slug that no longer
  resolves — logs and skips — because a deleted or renamed namespace must not
  cost somebody their login.
- Ships with the same preview-then-apply backfill, since adding a namespace
  after people have signed up leaves them all out.

Publishing only requires membership of any role, so being enrolled is enough
to publish there; no extra grant is needed.

Verified against a real PostgreSQL end to end: create a namespace, reject an
unknown slug, save with whitespace and duplicates and watch them normalise,
preview, apply, re-preview showing nothing left, and a fresh registration
landing in the global namespace, the new shared one, and its own personal one
at once.

As before, happy to split this out if you would rather keep the PR narrower.

FenjuFu:
Reviewed against #711 — the design lines up with what the issue asked for: operator-controlled, editable slug/name templates, off by default so existing deployments are unchanged on upgrade, and the "personal namespace = TEAM namespace with a single OWNER member" framing matches the current data model rather than inventing a visibility concept. The `AFTER_COMMIT` listener rationale is well argued.

Two blockers before review can proceed:

1. **DCO check is failing** — please sign off your commits (`git rebase --signoff` on the branch, then force-push).
2. **CLA is unsigned** — see the link on the `license/cla` check.

On the two-commit structure: keeping them in one PR is fine, the settings store reads better with its first consumer attached.

---

已对照 #711 核对：由运营方开关控制、slug/名称模板可配、默认关闭因此升级不影响存量部署，以及把"个人命名空间"落成"仅有一个 OWNER 成员的 TEAM 命名空间"而不新造可见性概念 —— 都符合 issue 的诉求，`AFTER_COMMIT` 监听器的理由也讲清楚了。

评审前有两个阻塞项：

1. **DCO 检查未通过** —— 请给提交补签名（分支上 `git rebase --signoff` 后 force-push）。
2. **CLA 未签署** —— 见 `license/cla` 检查中的链接。

关于两个 commit 的拆分：放在同一个 PR 里没问题，设置存储配上首个消费方更好判断。

FenjuFu (commented):
Reviewed the provisioning core against #711. The engineering here is solid — in particular the parts that are easy to get wrong are handled correctly:

- **After-commit transaction:** the listener fires on `@TransactionalEventListener` (AFTER_COMMIT), and `provisionFor` is annotated `@Transactional(propagation = REQUIRES_NEW)`. That is exactly right — a plain `@Transactional` there would join an already-committed context and silently drop the namespace write. Good catch.
- **Idempotency:** `if (alreadyOwnsNamespace(owner.userId())) return empty` guards both the listener path and the backfill against duplicate namespaces.
- **Collision + fail-open:** `MAX_SLUG_ATTEMPTS` slug allocation returns empty (logged) rather than throwing, and the listener swallows `RuntimeException`, so a naming clash or DB hiccup costs the user a namespace, not their registration/login. The backfill is batched with `MAX_BACKFILL_ACCOUNTS` and skips system/already-owning accounts.
- Operator-gated behind `settings.enabled()`, with admin settings + backfill endpoints.

**Blocking:** DCO is currently red (`ACTION_REQUIRED`) — the sign-off email on at least one commit doesn't match the commit author. Please `git commit --amend -s` (or rebase `--signoff`) so `Signed-off-by:` matches the author identity, and force-push.

This is a sizeable feature touching admin + namespace provisioning, so the go/no-go and rollout are the maintainers' call, but from a correctness standpoint it looks well-built. Once DCO is green it's in good shape for a maintainer pass.

XiaoSeS:
Maintainer scope review — please narrow this PR before merge:

The current head includes personal namespace provisioning, a generic `system_setting` store, default-namespace membership settings, backfill tooling, and the admin UI. Our accepted product scope is narrower: provision only newly activated users, use a stable random slug (`personal-<random>`), display name `${username}-个人空间`, keep existing users unchanged, and avoid a generic settings center/backfill in this PR.

Please either split the generic settings/default-membership/backfill work into follow-up PRs, or explain why those parts are required for the core behavior. Also add an explicit system-account guard inside the provisioning service (not only in backfill), and retain SUPER_ADMIN-only access for any remaining settings endpoint.

XiaoSeS:
Maintainer update pushed to the original head branch `feat/personal-namespace-provisioning` (commit `04bb414b`).

Applied: removed existing-user backfill and default-namespace membership expansion; default personal slug is now `personal-${random}` and display name `${username}-个人空间`; build passes.

Remaining follow-up in this same PR: replace the generic `system_setting` dependency with namespace-scoped settings (or settle that runtime admin editing is out of scope), and add an explicit system-account guard in provisioning. No merge recommendation until those are resolved.

PR #717, opened 2026-08-15
fix(web): avatar falls back to user initials when image fails to load

l3oferreira:
## Problem

When the user's `picture`/avatar claim from an OIDC identity provider points at an authenticated endpoint, the plain `` tag used for the avatar cannot load it, and the browser shows the broken-image icon (plus the `alt` text) instead of any avatar.

A concrete case: **Microsoft Entra ID** returns profile photos from `https://graph.microsoft.com/v1.0/me/photo/$value`, which requires a `Bearer` access token on the request. A browser-rendered `` has no way to attach that header, so the request is unauthenticated and fails — even though the `avatarUrl` value itself is valid and was correctly stored/returned by the backend.

This isn't specific to Entra: the same broken-image UI shows up for any IdP whose photo URL is not directly and anonymously fetchable, or simply for any avatar URL that has gone stale/404s.

## Before / after

- **Before**: header user menu, dashboard, and profile settings all rendered a raw ``. On load failure, the browser's broken-image icon + `alt` text showed instead of the user's photo.
- **After**: a new `UserAvatar` component renders the image, and on `onError` (or when there's no avatar URL at all) falls back to a small badge with the user's initials, styled to match the existing avatar sizing/shape at each call site (rounded-full in the header menu, rounded-2xl on the dashboard and profile pages).

## Change

- Add `web/src/shared/components/user-avatar.tsx`: a small component with `src`/`name`/`className`/`textClassName` props. It tracks an `onError` flag in state (reset when `src` changes) and swaps to a `div` with initials derived from `name` when there's no `src` or the image failed to load.
- Wire it into the three places that rendered a raw `` for the current user's avatar:
  - `web/src/shared/components/user-menu.tsx` (header/nav)
  - `web/src/pages/dashboard.tsx`
  - `web/src/pages/settings/profile.tsx`
- Add `web/src/shared/components/user-avatar.test.tsx` covering: image render, initials fallback with no `src`, initials fallback on `onError`, and initials derivation from a single-word name.

The project doesn't currently depend on `@radix-ui/react-avatar` (only `dialog`/`dropdown-menu`/`select` are used from Radix), so this adds a small local component instead of a new dependency, following the existing pattern of composed components under `web/src/shared/components/` (e.g. `namespace-badge.tsx`).

## Scope / non-goals

- No API or type changes — `avatarUrl` stays `string | undefined | null` everywhere.
- No new dependencies.
- Namespace avatars (`namespace-header.tsx`) are left untouched since they're not user photos and out of scope for this fix; happy to follow up if maintainers want the same treatment there.

## Testing

From `web/`:

```
pnpm install
pnpm lint       # eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0 -> clean
pnpm typecheck  # tsc --noEmit -> clean
pnpm test       # vitest run -> 193 files / 676 tests passed
pnpm build      # tsc -b && vite build -> succeeds
```

### DIFF
--- a/web/src/pages/dashboard.tsx
+++ b/web/src/pages/dashboard.tsx
@@ -7,6 +7,7 @@ import { canViewGovernanceCenter } from '@/shared/lib/governance-access'
 import { getHeadlineVersion } from '@/shared/lib/skill-lifecycle'
 import { TokenList } from '@/features/token/token-list'
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/shared/ui/card'
+import { UserAvatar } from '@/shared/components/user-avatar'
 import { APP_SHELL_PAGE_CLASS_NAME } from '@/app/page-shell-style'
 import { limitPreviewItems } from './dashboard-preview'
 
@@ -42,11 +43,12 @@ export function DashboardPage() {
         </CardHeader>
         <CardContent className="space-y-6">
           <div className="flex items-center gap-5">
-            {user?.avatarUrl && (
-              <img
+            {user && (
+              <UserAvatar
                 src={user.avatarUrl}
-                alt={user.displayName}
+                name={user.displayName}
                 className="h-20 w-20 rounded-2xl border-2 border-border/60 shadow-card"
+                textClassName="text-2xl font-semibold"
               />
             )}
             <div className="space-y-1.5">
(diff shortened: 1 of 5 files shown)


l3oferreira:
Closing — we'll carry this on our side for now. Feel free to reuse the patch if useful.

PR #724, opened 2026-08-18
feat(cli): add namespace workspace sync

15258725278:
## Summary

- add namespace-scoped `sync pull`, `status`, `diff`, and `push` workflows
- preserve public CLI 0.1.9 behavior while adding incremental workspace metadata and fingerprint checks
- bump the CLI package to 0.1.10
- document namespace workspace synchronization in the CLI README

## Verification

- `bun run lint`
- `bun run typecheck`
- `bun test` (389 passed)
- `bun run build`
- `node dist/index.js version` -> `SkillHub CLI 0.1.10`
- `node dist/index.js help sync`

The matching server namespace manifest and strict publish-validation APIs are already deployed and authenticated-smoke-tested in production.


FenjuFu (commented):
Reviewed the sync core. The fingerprinting and status model are clean and correct:

- `snapshotSkillDirectory` hashes each file with SHA-256 and folds sorted `path:hash` pairs into an aggregate `sha256:` fingerprint, so the workspace fingerprint is deterministic and order-independent.
- `status` does a proper three-way comparison — local directory snapshot vs stored workspace metadata vs remote — which lets it distinguish `local-changed` (snapshot ≠ metadata), `update-available` (metadata ≠ remote), `up-to-date`, and orphaned local dirs (flagging "local changes detected"). That is the right basis for a safe pull/push and mirrors git-style status.
- Unmanaged directories are called out with a distinct reason rather than being silently treated as installed.

Scope is contained to the CLI, existing 0.1.9 behavior is preserved, the version bump to 0.1.10 is included, and coverage is good (fingerprint/install-service/client unit tests plus the 187-line `sync-command` integration test; DCO green). One thing worth confirming in the description or a test: that `push` refuses to clobber when a slug is simultaneously `local-changed` **and** `update-available` (a true divergence), rather than pushing over newer remote state. Otherwise this looks well-built — merge decision to the maintainers.

PR #734, opened 2026-08-19
fix(web) 添加oidc的logo文件，配置oidc登录后图裂无logo太丑了

qwn3213:
image
如图


FenjuFu:
Thanks for adding the OIDC logo. One thing to double-check before this lands: the file is named `oidc-loge.svg`, which looks like a typo for `oidc-logo.svg`.

The other provider assets in `web/public` follow the `{provider}-logo.svg` convention (`github-logo.svg`, `gitlab-logo.svg`), so the login component almost certainly references `oidc-logo.svg`. If so, adding the file under the misspelled name won't actually fix the broken image — the `` would still 404. Could you confirm the filename exactly matches the `src` the OIDC login button uses, and rename to `oidc-logo.svg` if it follows the same convention? The SVG contents themselves look fine.

qwn3213:
Sorry, it was indeed a misnaming. It has been corrected.

FenjuFu:
Thanks — confirmed that the file is now correctly named `web/public/oidc-logo.svg`, so the filename issue is resolved.

The remaining blocker is DCO: both commits (`c421e08a` and `35488c6b`) currently have no `Signed-off-by` trailer. Please add a matching sign-off to every commit (for example, rebase the branch with `--signoff`), update the branch with `git push --force-with-lease`, and request re-review once the DCO check is green.

qwn3213:
DCO check is now green. Could you please re-review the changes?

PR #735, opened 2026-08-20
fix: emit confirm-publish events with permission-aware subscriber fanout

### LINKED ISSUE #731: [Bug] confirmPublish does not emit SkillPublishedEvent, so subscribers get no update notification
### Summary

`confirmPublish` transitions an `UPLOADED` version to `PUBLISHED`, but never publishes a `SkillPublishedEvent`. As a result, **users who subscribed to the skill receive no notification** when a new version goes live through this path.

The same "a new version is now available" outcome produces a notification when it happens via `SkillPublishService`, but is silent via `confirmPublish`. Subscribers cannot tell the difference — from their point of view the subscription simply stops working for some skills.

`SkillReviewSubmitService` already injects `ApplicationEventPublisher` (constructor line 47, field line 39), but never calls it — which suggests this is an oversight rather than an intentional design decision.

### Steps To Reproduce

1. User A publishes a skill to a team namespace with `visibility=PRIVATE`.
2. User B (a namespace member, not the author) subscribes to it — `PUT /api/v1/skills/{skillId}/subscription` returns `true`.
3. User A calls `POST /api/v1/skills/{ns}/{slug}/confirm-publish` with the new version.
4. Version status becomes `PUBLISHED`, `latestVersionId` is updated, download works.
5. **User B's `GET /api/v1/notifications/unread-count` stays at 0.**

Control case (works as expected):

- User A publishes with `visibility=PUBLIC` directly → subscriber B receives `PUBLISH | Skill updated: <slug>`.
- A version is yanked → subscriber B receives `PUBLISH | Skill version yanked: <slug>`.

Observed on `v0.2.16` (image revision `6e133c006e492dc3f468d91b21960aff1d577150`), self-hosted via `compose.release.yml`.

### Expected Behavior

`confirmPublish` should publish a `SkillPublishedEvent` after the version is persisted as `PUBLISHED`, so that `NotificationEventListener#onSkillPublishedForSubscribers` fans out to subscribers — consistent with the `autoPublish` branch in `SkillPublishService`.

### Root Cause

`server/skillhub-domain/src/main/java/com/iflytek/skillhub/domain/skill/service/SkillReviewSubmitService.java`

`confirmPublish` (line 113) ends with:

```java
// Update version to PUBLISHED
version.setStatus(SkillVersionStatus.PUBLISHED);
version.setPublishedAt(Instant.now(clock));
skillVersionRepository.save(version);

// Update skill's latest version
skill.setLatestVersionId(versionId);
skill.setUpdatedBy(actorUserId);
skillRepository.save(skill);
}   // <-- no event published
```

Compare with `SkillPublishService.java` line 571:

```java
if (autoPublish) {
    eventPublisher.publishEvent(new SkillPublishedEvent(skill.getId(), version.getId(), publisherId));
}
```

Occurrences of `eventPublisher.publishEvent`:

| Class | Count |
|---|---|
| `SkillPublishService` | 2 |
| `SkillReviewSubmitService` | **0** |

`NotificationEventListener#onSkillPublishedForSubscribers` (which calls `skillSubscriptionService.findSubscribersBySkillId`) is therefore never reached from this path. The subscriber fan-out logic itself is correct and works — it is only the event emission that is missing.

### Suggested Fix

Append to the end of `confirmPublish`:

```java
eventPublisher.publishEvent(new SkillPublishedEvent(skill.getId(), versionId, actorUserId));
```

One consideration worth confirming by a maintainer: `confirmPublish` requires `visibility == PRIVATE`, so subscribers who are not namespace members must not be able to infer content they cannot access. Since the notification only carries the skill name and version label (no package content), and subscription itself already requires read access to the skill, this appears safe — but the fan-out may still want to filter recipients by current read permission, which would also harden the existing `PUBLIC` path.

### Impact

Teams that follow the recommended governance flow — publish as `PRIVATE`, verify, then confirm — get no update notifications at all, which is precisely the workflow where dependency tracking matters most. Consumers silently keep using a stale version.

Happy to submit a PR if the suggested approach (plus an optional recipient-permission filter) looks right.

xiaoyaner0201:
## Summary

- emit `SkillPublishedEvent` after `confirmPublish` persists the published version and latest skill version
- apply purpose-aware, current metadata-read eligibility before publish/yank subscriber fanout
- fail closed on stale membership, inactive accounts/namespaces, cross-namespace state, or authoritative lookup failures

Fixes #731.

## Why

The confirm-publish path currently updates persistence without reaching the existing subscriber listener. For PRIVATE skills, fanout also needs to re-check current read eligibility so removed or stale subscribers do not receive recipient-visible metadata.

## Validation

- [x] Backend tests passed
- [ ] Frontend typecheck/build passed — not applicable; no frontend files changed
- [x] OpenAPI SDK regenerated or checked when API contracts changed — not applicable; no API contract changed
- [ ] Smoke test run when relevant — not run; no operator-facing workflow changed

Commands and results:

```text
exact listener → eligibility → dispatcher → Notification persistence/SSE sink probe: 8 passed
authorization and compatibility suite: domain 76, notification 4, app 49; 0 failures/errors/skips
full reactor: 1,400 tests, 0 failures/errors; 1 existing Redis-cluster environment skip
git diff --check: passed
```

## Risk

- User-facing impact: subscribers receive confirm-publish and yank notifications only while they retain current metadata-read eligibility
- Deployment or migration impact: none; no schema or API migration
- Rollback approach: revert this commit to restore the previous event and fanout behavior

## Limits

Not exercised against production infrastructure, a real browser EventSource client, or a configured Redis cluster. The synchronous production listener/eligibility/dispatcher path and final persistence/SSE arguments are covered.

## Notes

- Related issue: Fixes #731
- This is opened as a draft because the change crosses authorization and notification fanout boundaries and should receive focused maintainer review.

### DIFF
--- a/server/skillhub-app/src/main/java/com/iflytek/skillhub/listener/NotificationEventListener.java
+++ b/server/skillhub-app/src/main/java/com/iflytek/skillhub/listener/NotificationEventListener.java
@@ -8,6 +8,7 @@ import com.iflytek.skillhub.domain.skill.Skill;
 import com.iflytek.skillhub.domain.skill.SkillRepository;
 import com.iflytek.skillhub.domain.skill.SkillVersionRepository;
 import com.iflytek.skillhub.domain.social.SkillSubscriptionService;
+import com.iflytek.skillhub.domain.social.SubscriptionRecipientEligibility;
 import com.iflytek.skillhub.notification.domain.NotificationCategory;
 import com.iflytek.skillhub.notification.service.NotificationDispatcher;
 import org.slf4j.Logger;
@@ -33,6 +34,7 @@ public class NotificationEventListener {
     private final NotificationDispatcher dispatcher;
     private final SkillSubscriptionService skillSubscriptionService;
     private final ObjectMapper objectMapper;
+    private final SubscriptionRecipientEligibility subscriptionEligibility;
 
     public NotificationEventListener(SkillRepository skillRepository,
                                       SkillVersionRepository skillVersionRepository,
@@ -40,7 +42,8 @@ public class NotificationEventListener {
                                       RecipientResolver recipientResolver,
                                       NotificationDispatcher dispatcher,
                                       SkillSubscriptionService skillSubscriptionService,
-                                      ObjectMapper objectMapper) {
+                                      ObjectMapper objectMapper,
+                                      SubscriptionRecipientEligibility subscriptionEligibility) {
         this.skillRepository = skillRepository;
         this.skillVersionRepository = skillVersionRepository;
         this.namespaceRepository = namespaceRepository;
@@ -48,6 +51,7 @@ public class NotificationEventListener {
         this.dispatcher = dispatcher;
         this.skillSubscriptionService = skillSubscriptionService;
         this.objectMapper = objectMapper;
+        this.subscriptionEligibility = subscriptionEligibility;
     }
 
     @Async("skillhubEventExecutor")
@@ -74,6 +78,8 @@ public class NotificationEventListener {
             if (subscribers.isEmpty()) {
                 return;
             }
+            var namespace = namespaceRepository.findById(skill.getNamespaceId()).orElse(null);
+            subscribers = subscriptionEligibility.currentRecipients(skill, namespace, subscribers);
             String title = "Skill updated: " + skillDisplayName(skill);
             Map<String, Object> body = bodyWithSkill(skill);
             versionLabel(event.versionId(), body);
@@ -96,6 +102,8 @@ public class NotificationEventListener {
             if (subscribers.isEmpty()) {
                 return;
             }
+            var namespace = namespaceRepository.findById(skill.getNamespaceId()).orElse(null);
+            subscribers = subscriptionEligibility.yankedRecipients(skill, namespace, subscribers, event.wasPublished());
             String title = "Skill version yanked: " + skillDisplayName(skill);
             Map<String, Object> body = bodyWithSkill(skill);
             versionLabel(event.versionId(), body);
--- a/server/skillhub-app/src/test/java/com/iflytek/skillhub/domain/social/SkillSubscriptionServiceTest.java
+++ b/server/skillhub-app/src/test/java/com/iflytek/skillhub/domain/social/SkillSubscriptionServiceTest.java
@@ -2,6 +2,17 @@ package com.iflytek.skillhub.domain.social;
 
 import com.iflytek.skillhub.domain.skill.Skill;
 import com.iflytek.skillhub.domain.skill.SkillRepository;
+import com.iflytek.skillhub.domain.namespace.Namespace;
+import com.iflytek.skillhub.domain.namespace.NamespaceMember;
+import com.iflytek.skillhub.domain.namespace.NamespaceMemberRepository;
+import com.iflytek.skillhub.domain.namespace.NamespaceRepository;
+import com.iflytek.skillhub.domain.namespace.NamespaceRole;
+import com.iflytek.skillhub.domain.namespace.NamespaceStatus;
+import com.iflytek.skillhub.domain.skill.SkillVisibility;
+import com.iflytek.skillhub.domain.user.UserAccount;
+import com.iflytek.skillhub.domain.user.UserAccountRepository;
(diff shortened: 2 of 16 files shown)


FenjuFu (commented):
Deep-reviewed against #731 and the repo's visibility model. The core fix is correct and the extra hardening is well thought through — a few actionable points below.

**Core fix (resolves #731):** `confirmPublish` now publishes `SkillPublishedEvent` after persisting the PUBLISHED version and `latestVersionId`, so `NotificationEventListener#onSkillPublishedForSubscribers` is finally reached — consistent with the `autoPublish` branch in `SkillPublishService`. 

**On the recipient filtering:** I initially flagged that the issue's scenario (a regular namespace *member* subscribing to a PRIVATE skill and expecting a notification) is at odds with `SubscriptionMetadataAccessPolicy` returning `owner || manager` for PRIVATE. After checking `docs/oss-02-core-semantic-rules.md`, the canonical read model is `PRIVATE -> owner only, not marketplace-visible`, so the original "member gets notified" behavior was itself fail-open. Filtering non-eligible subscribers is the correct call, and allowing `manager` (ADMIN/OWNER) is a reasonable governance extension. The batch account/member loading (`findByIdIn` / `findByNamespaceIdAndUserIdIn`) avoids an N+1 in the async fan-out. Good.

Points worth addressing / a maintainer decision:

1. **Missing i18n key.** `subscribe()` throws `new DomainForbiddenException("error.skill.subscription.noPermission")`, but I don't see that key added to `messages.properties` / `messages_zh.properties` (the diff is Java-only). `ApiResponseFactory` resolves via `getMessage(code, args, code, locale)`, so this degrades to showing the raw code rather than a 500 — not fatal, but existing forbidden keys (e.g. `error.namespace.membership.required`) are registered, so please add EN/zh entries for consistency.

2. **`subscribe()` is now a behavior change on a public endpoint.** Previously `PUT /skills/{id}/subscription` succeeded for any existing skill; it now returns 403 for users without current read access, and existing regular-member subscriptions to PRIVATE/NAMESPACE_ONLY skills will be silently filtered out of all future notifications. That's the right hardening, but it's a maintainer call whether a one-time prune/migration of now-ineligible `skill_subscription` rows is wanted so the DB state matches the new policy.

3. **Fail-closed on a null namespace.** In the listener, `namespaceRepository.findById(...).orElse(null)` feeding an eligibility check that denies when `namespace == null` means a namespace-load hiccup silently drops the whole fan-out. Safe by default, but a debug/warn log there would make such a case diagnosable rather than invisible.

Only DCO has reported so far — worth waiting for the backend build/test job before merge so the new listener/policy tests are confirmed green in CI. Correctness and design look right to me; leaving the merge and the policy/migration decision to the maintainers.

XiaoSeS:
Thanks for the PR. The narrow #731 regression has now been fixed by #738, which is merged into main.

The broader permission-aware subscriber fanout in this PR is still valuable, but it should be handled as a separate authorization hardening change rather than bundled with the confirmPublish event fix.

Please keep this PR open as draft/design work, rebase it on latest main, remove the already-merged confirmPublish change, and narrow the scope to:

1. reject subscription creation when the current user cannot read the skill metadata;
2. filter subscriber notifications using the current visibility/access rules at fanout time;
3. reuse the existing VisibilityChecker or a shared access policy instead of introducing a parallel visibility model;
4. add the missing i18n message and tests for private, namespace-only, hidden, removed-member, and disabled-user cases.

We should not close this PR, but it is not mergeable as-is.

XiaoSeS:
维护者已按评审意见创建跟进 PR #765（https://github.com/iflytek/skillhub/pull/765）。该跟进已移除已合并的 #731 confirmPublish 变更，范围收窄为订阅创建鉴权、发布/下架通知扇出时的当前可见性过滤、复用 VisibilityChecker、补充中英文消息和回归测试。原 PR 保持开放 Draft，待 #765 验证合并后再决定是否关闭。

XiaoSeS:
权限感知订阅扇出已由维护者跟进 PR #765 修复并合并（merge commit `2e78f79e836cecc9f8f0947a4badb83bb351da7f`）。原 PR 的 Draft/Dirty 版本不再继续，现关闭以避免重复实现。

PR #747, opened 2026-08-24
fix(dashboard): remove encoding from slug params

myml:
## What

Remove explicit encoding from slug parameters in dashboard page.

## Why

The router handles parameter encoding automatically. Explicitly encoding the slug results in double-encoded characters in the URL.

## How

Remove the `encodeURIComponent` call when passing slug to router navigation.

## Testing

- Navigate to dashboard and click on skill items
- Verify URL parameters are correctly encoded (not double-encoded)
- Test with Chinese characters in skill names

## Impact

- Fixes URL encoding issue for skills with special characters in names
- No breaking changes

### DIFF
--- a/web/src/pages/dashboard.tsx
+++ b/web/src/pages/dashboard.tsx
@@ -144,7 +144,7 @@ export function DashboardPage() {
                       <Link
                         key={skill.id}
                         to="/space/$namespace/$slug"
-                        params={{ namespace: skill.namespace, slug: encodeURIComponent(skill.slug) }}
+                        params={{ namespace: skill.namespace, slug: skill.slug }}
                         className="rounded-lg border border-border/60 px-3 py-3 transition-colors hover:bg-accent/40"
                       >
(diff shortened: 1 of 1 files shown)


FenjuFu:
Correct fix — TanStack Router percent-encodes path params when it builds the URL, so the explicit `encodeURIComponent` was double-encoding (visible as `%25XX` for non-ASCII slugs, same class of bug as the ClawHub redirect encoding issue). Low risk. Just confirm the `$slug` route reads the param already-decoded (router default) so the detail page still matches — a quick click-through on a Chinese-slug skill from the dashboard covers it.

myml:
Spot on – double‑encoding fixed. Verified with a Chinese‑slug skill: URL is clean, detail page renders fine. Ready to go, resolving.

FenjuFu (approved):
Re-reviewed the one-line fix and the reported Chinese-slug click-through. TanStack Router owns path-param encoding, so passing the raw slug removes the double-encoding without changing route semantics. All remote checks, DCO, and CLA are green.

Issue #622, opened 2026-07-29
[Feature] Replace notification SSE with proxy-friendly HTTP polling


XiaoSeS:
### Problem

The notification badge and list currently depend on a long-lived SSE connection. Although the current implementation has reconnect and heartbeat handling, SSE still adds a separate server-side emitter lifecycle, async MVC timeout configuration, frontend connection coordination, and reverse-proxy requirements that are disproportionate to the notification UI's eventual-consistency needs.

Self-hosted deployments commonly sit behind Nginx, Ingress, or other proxies where buffering, idle timeouts, connection limits, and reloads can make long-lived streams harder to operate than ordinary authenticated HTTP requests. The existing notification list and unread-count endpoints already provide all data needed for a polling transport.

### Proposed Solution

Replace the web notification SSE path with bounded HTTP polling as one complete slice:

- [ ] Poll the authenticated unread-count and visible notification-list queries on a fixed interval, with an immediate refresh when the window regains focus.
- [ ] Keep polling scoped to the active authenticated user and stop it after logout or when no user session exists.
- [ ] Preserve immediate local cache updates after mark-read, mark-all-read, and delete actions instead of waiting for the next interval.
- [ ] Remove the frontend EventSource coordinator and the backend emitter/heartbeat lifecycle once polling coverage is in place.
- [ ] Remove the SSE endpoint and SSE-specific async timeout/error handling, then regenerate the OpenAPI web types.
- [ ] Document the delivery latency and request-volume tradeoff for self-hosted operators.
- [ ] Add frontend tests for interval refresh, focus refresh, mutation cache behavior, session changes, and cleanup; add backend regression tests for the remaining list/unread endpoints.

A 10-second default interval is sufficient for the current in-app notification use case and avoids presenting polling as real-time delivery.

### Alternatives Considered

- Keep SSE only: provides lower delivery latency, but retains proxy-specific operational requirements and substantially more connection-management code.
- Add polling only as an automatic SSE fallback: supports both environments but keeps both transports, duplicate cache paths, and more states to test.
- Use WebSockets: increases infrastructure and protocol complexity without a current bidirectional messaging requirement.

### Impact

The change simplifies backend and frontend notification delivery but increases periodic HTTP requests while an authenticated page is open. Notification delivery becomes eventually consistent within the polling interval rather than near-real-time. Existing notification persistence, preferences, pagination, unread counts, and mutation behavior should remain unchanged.

[...]

PR #748, opened 2026-08-24
fix(security): harden CSP policy and hide nginx server tokens

myml:
## What

Harden Content Security Policy (CSP) and hide nginx server version information.

## Why

**CSP Hardening:**
- `'unsafe-inline'` and `'unsafe-eval'` in script-src allow inline scripts and eval(), which can enable XSS attacks
- `ws: wss: http://localhost:* https://localhost:*` in connect-src allows WebSocket and localhost connections that may not be needed in production

**Nginx Server Tokens:**
- By default, nginx exposes its version in HTTP response headers (`Server: nginx/x.x.x`)
- This information can help attackers identify known vulnerabilities

## How

1. Remove `'unsafe-inline' 'unsafe-eval'` from script-src directive
2. Restrict connect-src to `'self'` only
3. Add `server_tokens off;` to nginx configuration

## Testing

- Verify the application still works correctly with the stricter CSP
- Check nginx response headers do not contain version information

## Impact

- May break functionality that relies on inline scripts or eval() - needs testing
- WebSocket connections from frontend will be blocked - verify if this is intentional
- Nginx version will no longer be exposed in Server header

### DIFF
--- a/server/skillhub-auth/src/main/java/com/iflytek/skillhub/auth/config/SecurityConfig.java
+++ b/server/skillhub-auth/src/main/java/com/iflytek/skillhub/auth/config/SecurityConfig.java
@@ -45,11 +45,11 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
 public class SecurityConfig {
     private static final String CONTENT_SECURITY_POLICY = String.join("; ",
             "default-src 'self'",
-            "script-src 'self' 'unsafe-inline' 'unsafe-eval'",
+            "script-src 'self'",
             "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com",
             "img-src 'self' data: blob: https:",
             "font-src 'self' data: https://fonts.gstatic.com",
-            "connect-src 'self' ws: wss: http://localhost:* https://localhost:*",
+            "connect-src 'self'",
             "object-src 'none'",
             "base-uri 'self'",
             "frame-ancestors 'none'",
--- a/web/nginx.conf.template
+++ b/web/nginx.conf.template
@@ -1,3 +1,4 @@
+server_tokens off;
 server {
     listen 80;
     server_name _;


FenjuFu:
The direction is right (dropping `'unsafe-inline'`/`'unsafe-eval'` from `script-src` and tightening `connect-src` are real XSS-surface reductions, and `server_tokens off` is a clean win). The risk is that these three are the classic ways a strict CSP breaks a running SPA, so this needs a smoke test against the **production build** (not dev) before merge:

- **`'unsafe-eval'` removal**: breaks any dependency that uses `eval`/`new Function` (some date/i18n/wasm/templating libs do). Worth a grep of the bundle.
- **`'unsafe-inline'` removal from `script-src`**: breaks if `index.html` or any lib injects an inline `<script>` or inline event handlers. Vite's module output is external, so usually fine — but verify.
- **`connect-src 'self'`**: drops `ws:`/`wss:` and `localhost:*`. Same-origin SSE/fetch stay allowed, but if notifications still use a WebSocket or any cross-origin endpoint (and #622 is still tracking the SSE→polling move), they'll be blocked. Confirm every XHR/fetch/SSE target is same-origin.

If you've already run the built frontend through login + notifications + skill browse with the console open and it's clean, say so in the PR and this is good to go.

myml:
We've run through login, notifications, and skill browsing on the built frontend with the console open – no CSP errors and all features work as expected. This has been running stably in our internal production for several days, so it should be good to merge. @FenjuFu

FenjuFu (approved):
Re-reviewed after the production-build smoke evidence. The diff is limited to removing script unsafe-inline and unsafe-eval, restricting connect-src to same-origin, and disabling nginx server tokens. Login, notifications, and skill browsing were exercised without CSP violations, and all remote checks, DCO, and CLA are green.

XiaoSeS:
Thanks for confirming the production-build smoke coverage for login, notifications, and skill browsing. That addresses the main runtime-breakage concern raised earlier.

One remaining implementation detail to clarify before merge: the PR tightens CSP in `SecurityConfig`, but `web/index.html` still contains the older CSP meta policy:

```html
script-src 'self' 'unsafe-inline' 'unsafe-eval';
connect-src 'self' ws: wss: http://localhost:* https://localhost:*;
```

In the release deployment, the SPA entry page is served directly by nginx, so the backend `SecurityConfig` CSP mainly applies to backend/API responses and may not be the effective policy for the Web entry page. This can leave the project with two different CSP sources and make the actual browser policy deployment-dependent.

Suggested adjustment:

1. Make the Web CSP authority explicit, preferably at the nginx/static-web boundary, or keep `web/index.html` synchronized if the meta tag remains.
2. Re-check whether `connect-src 'self'` is compatible with deployments using `SKILLHUB_WEB_API_BASE_URL` for a non-same-origin API. If same-origin API is the intended production contract, documenting that assumption is enough.
3. `server_tokens off;` looks good and can stay as-is.

After that clarification/fix plus the already mentioned production-build smoke, this should be safe to merge.

PR #751, opened 2026-08-24
feat(skill-card): show author and update time in skill card

myml:
## What

Add author and last update time display at the bottom of skill cards.

## Why

Users want to see who created a skill and when it was last updated directly in the search results, without opening the detail page.

## How

1. Add `ownerId` and `ownerDisplayName` fields to `SkillSummaryResponse`
2. Update `JpaMySkillQueryRepository` and `SkillSearchAppService` to include owner info
3. Update frontend types and `SkillCard` component to display author and update time

## Testing

- Verify skill cards show author name (when available)
- Verify skill cards show relative update time (e.g., "2天前")
- Verify backward compatibility when owner info is not available

## Impact

- New API fields: `ownerId`, `ownerDisplayName` in `SkillSummaryResponse`
- UI change: author and update time shown at the bottom of skill cards
- No breaking changes

### DIFF
--- a/server/skillhub-app/src/main/java/com/iflytek/skillhub/dto/SkillSummaryResponse.java
+++ b/server/skillhub-app/src/main/java/com/iflytek/skillhub/dto/SkillSummaryResponse.java
@@ -16,6 +16,8 @@ public record SkillSummaryResponse(
         Integer ratingCount,
         String namespace,
         Instant updatedAt,
+        String ownerId,
+        String ownerDisplayName,
         boolean canSubmitPromotion,
         SkillLifecycleVersionResponse headlineVersion,
         SkillLifecycleVersionResponse publishedVersion,
--- a/server/skillhub-app/src/main/java/com/iflytek/skillhub/repository/JpaMySkillQueryRepository.java
+++ b/server/skillhub-app/src/main/java/com/iflytek/skillhub/repository/JpaMySkillQueryRepository.java
@@ -75,6 +75,8 @@ public class JpaMySkillQueryRepository implements MySkillQueryRepository {
                 skill.getRatingCount(),
                 namespace != null ? namespace.getSlug() : null,
                 skill.getUpdatedAt(),
+                skill.getOwnerId(),
+                null,
                 canSubmitPromotion(skill, publishedVersion, namespace),
                 toLifecycleVersion(headlineVersion),
                 toLifecycleVersion(publishedVersion),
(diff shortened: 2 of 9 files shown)


FenjuFu:
The "update time" half is fully wired, but the **author name is never populated**: both mappers (`JpaMySkillQueryRepository.toSummaryResponse` and `SkillSearchAppService.toSummaryResponse`) pass `ownerDisplayName = null`, so the new field is always null and the card can only render the raw `ownerId`. To actually "show author" you need to resolve `ownerId → displayName` — ideally a batch user lookup over the page's owner IDs (avoid a per-card N+1). As written, the feature ships the plumbing but not the display value.

XiaoSeS:
Thanks for the PR. The existing review comment already covers the main blocker: `ownerDisplayName` is added to the response type, but both backend mappers still pass `null`, so the card cannot actually show the author name yet.

One additional point: the new relative-time formatter in `skill-card.tsx` hardcodes Chinese strings (`刚刚`, `分钟前`, etc.). SkillHub already supports multiple UI languages, so this should either use the existing i18n pattern or extract a shared relative-time helper that accepts the active language. Otherwise this feature will regress non-Chinese locales on skill cards.

Suggested path:

1. Populate `ownerDisplayName` from `ownerId` without a per-card N+1 lookup, preferably by batching owner IDs at the query/application-service boundary.
2. Add backend tests proving the API returns the display name when available and has a clear fallback when it is not.
3. Replace the hardcoded Chinese formatter with an i18n-aware/shared formatter, and cover at least the default locale behavior.

If you want to keep this PR smaller, an acceptable alternative is to scope it down to “show updated time only” and remove the unused owner fields for now.

XiaoSeS:
Review finding: the new `ownerDisplayName` field is always serialized as `null` in both `JpaMySkillQueryRepository` and `SkillSearchAppService`, while the web card only renders the author when `ownerDisplayName` is present. As a result, the PR currently delivers the update-time display but not the author display claimed by its title. Please either batch-load owner display names from the existing user repository (avoid N+1 queries) or narrow the PR title/scope to update time only. Add a regression assertion for the search/my-skills summary author field before requesting re-review.

XiaoSeS:
维护者复核发现：原 PR 已增加 `ownerDisplayName` 字段，但查询层始终返回 `null`，因此卡片无法显示作者。已创建维护者跟进 PR #764（https://github.com/iflytek/skillhub/pull/764），通过批量读取用户资料补齐作者显示，并增加回归断言。原 PR 暂不关闭，待 #764 合并后再统一处理。

XiaoSeS:
已由维护者跟进 PR #764 修复并合并（merge commit `7476c9e0d20bb8c941fa4d36bb41be01232b7afc`）：作者显示名已通过批量查询补齐，并保留兼容构造函数。现关闭此原始 PR。

PR #752, opened 2026-08-24
fix(scan): prevent scan tasks from being lost after transaction commit

### LINKED ISSUE #749: [Bug] 扫描任务可能在事务提交后丢失
## Description

扫描流程先将 SkillVersion 更新为 `SCANNING`，事务提交后再写入 Redis Stream。若数据库提交成功后，应用在 Redis 写入前崩溃，或者 Redis 暂时不可用，扫描任务将不会进入队列。

当前消费者只能恢复已经写入 Redis Pending List 的消息，无法恢复从未成功写入 Redis 的任务。

## Effect

SkillVersion 会永久停留在 `SCANNING` 状态，无法继续完成安全扫描，也无法进入后续的 `UPLOADED` 或 `PENDING_REVIEW` 状态。

用户可能看到上传已完成，但技能长期不可用；管理员也只能通过人工修改数据库或重新触发扫描恢复。

实际结果：Redis 中没有对应任务，SkillVersion 仍为 `SCANNING`。

## Reproduction

1. 提交一个需要安全扫描的 Skill。
2. 确认数据库事务已提交，SkillVersion 状态为 `SCANNING`。
3. 在 Redis Stream 写入前停止应用，或使 Redis 暂时不可用。
4. 恢复应用和 Redis。
5. 检查 Redis Stream 及 SkillVersion 状态。

### Expected Behavior

数据库提交成功后，即使应用重启或 Redis 暂时不可用，扫描任务也应能够自动补发，最终完成扫描或明确进入失败状态。

### Environment

_No response_

### API Contract Impact

_No response_

### Logs Or Screenshots

_No response_

bbdu3:
## Summary

This closes #749 .Fix the issue where scan tasks can be lost after the database transaction commits but before they are written to the Redis Stream.

## Changes

- Added the `scan_task_outbox` table, domain model, and repository implementation.
- Persist security audit records and Outbox tasks within the same transaction.
- Added a scheduled Dispatcher supporting:
  - Publishing tasks to the Redis Stream;
  - Retry with exponential backoff;
  - Reclaiming timed-out tasks;
  - Cleaning up completed tasks.
- Added `taskId`-based consumer idempotency checks to prevent duplicate scans.
- Clean up related Outbox tasks when a `SkillVersion` is deleted.
- Added tests for Outbox state transitions, transactional persistence, and Redis publish failures.

## Verification

- `SecurityScanServiceTest`: 10 passed
- `ScanTaskOutboxTest`: 2 passed
- `SecurityScanOutboxTest`: 1 passed
- `skillhub-infra` compilation passed
- `git diff --check` passed

### DIFF
--- a/server/skillhub-app/src/main/java/com/iflytek/skillhub/stream/ScanTaskConsumer.java
+++ b/server/skillhub-app/src/main/java/com/iflytek/skillhub/stream/ScanTaskConsumer.java
@@ -11,6 +11,7 @@ import com.iflytek.skillhub.domain.skill.SkillVersionRepository;
 import com.iflytek.skillhub.domain.skill.SkillVersionStatus;
 import com.iflytek.skillhub.observability.MessageObservationSupport;
 import com.iflytek.skillhub.storage.ObjectStorageService;
+import org.redisson.api.RLock;
 import org.redisson.api.RedissonClient;
 
 import java.io.IOException;
@@ -26,6 +27,7 @@ import java.util.Map;
 public class ScanTaskConsumer extends AbstractStreamConsumer<ScanTaskConsumer.ScanTaskPayload> {
     private static final Path SCAN_TEMP_DIR = Paths.get("/tmp/skillhub-scans").toAbsolutePath().normalize();
 
+    private final RedissonClient redissonClient;
     private final SecurityScanner securityScanner;
     private final SecurityScanService securityScanService;
     private final SkillVersionRepository skillVersionRepository;
@@ -42,6 +44,7 @@ public class ScanTaskConsumer extends AbstractStreamConsumer<ScanTaskConsumer.Sc
                             ObjectStorageService objectStorageService,
                             MessageObservationSupport messageObservationSupport) {
         super(redissonClient, streamKey, groupName, messageObservationSupport);
+        this.redissonClient = redissonClient;
         this.securityScanner = securityScanner;
         this.securityScanService = securityScanService;
         this.skillVersionRepository = skillVersionRepository;
@@ -72,6 +75,7 @@ public class ScanTaskConsumer extends AbstractStreamConsumer<ScanTaskConsumer.Sc
                 reclaimInterval,
                 messageObservationSupport
         );
+        this.redissonClient = redissonClient;
         this.securityScanner = securityScanner;
         this.securityScanService = securityScanService;
         this.skillVersionRepository = skillVersionRepository;
@@ -132,13 +136,31 @@ public class ScanTaskConsumer extends AbstractStreamConsumer<ScanTaskConsumer.Sc
             log.info("Skipping already processed security scan task: taskId={}, versionId={}", payload.taskId(), payload.versionId());
             return;
         }
+        RLock processingLock = redissonClient.getLock("skillhub:scan:processing:" + payload.taskId());
+        boolean acquired = false;
+        try {
+            acquired = processingLock.tryLock();
+            if (!acquired) {
+                log.info("Skipping concurrently processed security scan task: taskId={}, versionId={}",
+                        payload.taskId(), payload.versionId());
+                payload.skipCleanup();
+                return;
+            }
+            if (securityScanService.isTaskAlreadyProcessed(payload.taskId())) {
+                return;
+            }
+            executeScan(payload);
+        } finally {
+            if (acquired && processingLock.isHeldByCurrentThread()) {
+                processingLock.unlock();
+            }
+        }
+    }
+
+    private void executeScan(ScanTaskPayload payload) {
         String skillPath = resolveWorkingSkillPath(payload);
         SecurityScanRequest request = new SecurityScanRequest(
-                payload.taskId(),
-                payload.versionId(),
-                skillPath,
-                Map.of()
-        );
+                payload.taskId(), payload.versionId(), skillPath, Map.of());
         SecurityScanResponse response = securityScanner.scan(request);
         securityScanService.processScanResult(payload.versionId(), payload.scannerType(), response);
     }
@@ -263,6 +285,7 @@ public class ScanTaskConsumer extends AbstractStreamConsumer<ScanTaskConsumer.Sc
         private final ScannerType scannerType;
         private final int retryCount;
         private String workingSkillPath;
+        private boolean cleanupEnabled = true;
 
         protected ScanTaskPayload(String taskId, Long versionId, String skillPath, String bundleKey, ScannerType scannerType) {
             this(taskId, versionId, skillPath, bundleKey, scannerType, 0);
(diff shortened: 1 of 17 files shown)


FenjuFu:
Reviewed against #749. The transactional-outbox approach is the right fix: writing `ScanTaskOutbox(PENDING)` inside the same transaction as the `SkillVersion -> SCANNING` / `SecurityAudit(taskId)` update makes enqueue atomic with the state change, so a crash or Redis outage between commit and publish can no longer strand a version in `SCANNING`. The dispatcher's claim/lease + expired-lease re-claim, the exponential backoff, the SENT cleanup job, and the consumer-side idempotency guard are all coherent, and the test coverage is good. A few points, most tied back to the issue's own acceptance wording:

1. **`FAILED` is defined but never reached — the issue's "明确进入失败状态" outcome isn't implemented.** The issue's Expected Behavior is "最终完成扫描或**明确进入失败状态**". `markRetry` always returns the row to `PENDING`; there is no max-retry cap that transitions to `FAILED`. A permanently-broken task (bad payload, scanner rejects it) will re-publish every `max-backoff` (5 min) forever and never surface to an admin. Consider a `max-attempts` threshold that flips the row to `FAILED` (and ideally drives the `SkillVersion` out of `SCANNING` into a failed state) so the "clearly failed" half of the issue is actually met.

2. **Multi-replica dispatch aborts the whole batch on the first claim collision.** `findPendingDue`/`findExpiredLeases` are plain selects (no `FOR UPDATE SKIP LOCKED`), and cross-instance safety rests on the `@Version` optimistic lock at `saveAndFlush(outbox)`. That `saveAndFlush` is outside the per-task `try`, so if two dispatchers claim the same row, the `OptimisticLockException` propagates out of the `@Transactional dispatch()` and rolls back the entire batch (every already-claimed row in that run reverts). It self-heals next tick, but under contention it can thrash. Either catch the optimistic failure per row and `continue`, or switch the two finders to `SKIP LOCKED` pessimistic locking.

3. **Idempotency guard only covers *completed* scans, not in-flight ones.** `isTaskAlreadyProcessed` = `existsByTaskIdAndScannedAtIsNotNull`. At-least-once redelivery (publish succeeds, then the app dies before `markSent`, lease expires, re-published) is possible by design — but if the redelivery lands before the first scan finishes, both pass the guard and the scanner runs twice concurrently for one `taskId`. If `markProcessing` doesn't already dedup in-flight tasks, consider guarding on "audit exists for taskId" rather than "audit scanned".

4. **`softDeleteByVersionId` doesn't purge the outbox (only `hardDelete` does).** A version soft-deleted while a `PENDING` outbox row is still queued will still get published and scanned by the dispatcher. Mirror the `scanTaskOutboxRepository.deleteByVersionId(...)` call (or mark the row terminal) in the soft-delete path too.

5. **Minor:** `ScanTaskOutbox.toScanTask()` reconstructs the payload from discrete columns and hardcodes `scannerType = SKILL_SCANNER`, dropping the original `ScanTask` attributes map — fine today since `triggerScan` only ever emits `SKILL_SCANNER`, but it silently won't round-trip any future scanner type/attribute. Also the entity's constructor/`@PrePersist` use `Instant.now(Clock.systemUTC())` directly while the dispatcher uses the injected `Clock` — worth unifying on the injected clock for testability.

None of these block the core fix for the strand-in-SCANNING bug; (1) is the one I'd most want addressed since it's part of the issue's stated expected behavior.

bbdu3:
Thanks for the detailed review. I’ve addressed all five points:

1. Added a configurable maximum attempt limit. Exhausted tasks now move to `FAILED`, and versions still in `SCANNING` move to `SCAN_FAILED`.
2. Changed dispatcher selection to PostgreSQL `FOR UPDATE SKIP LOCKED` to prevent multi-replica claim collisions from rolling back an entire batch.
3. Added a Redisson watchdog lock keyed by `taskId` to prevent concurrent duplicate scans. Also fixed edge case where a skipped duplicate could delete the active scan’s temporary directory.
4. Soft deletion now removes pending Outbox tasks for the corresponding version.
5. Added a `V45` migration to preserve the original task metadata and creation time instead of reconstructing a partial payload.
Additional tests cover retry exhaustion, terminal-state protection, expired lease reclaiming, concurrent delivery, lock release on failure, soft deletion, and payload round-tripping.

Verification:
- Domain tests: 15 passed
- Consumer/Dispatcher/Logging tests: 18 passed
- Full backend suite: 1,392 tests, 0 failures, 0 errors, 1 skipped
- Clean Java 21 build: all 8 Maven modules passed
- `git diff --check` passed

FenjuFu (dismissed):
The five requested changes are present, but the new taskId lock introduces a message-loss race with the existing pending-message reclaimer.\n\nIn ScanTaskConsumer.processBusiness, a failed processingLock.tryLock() calls skipCleanup() and returns normally. AbstractStreamConsumer.handleMessageInScope treats every normal return as success: it calls markCompleted(payload) and then cknowledge(messageId).\n\nThat is unsafe for the exact duplicate-delivery scenario this lock targets:\n\n1. Consumer A is still scanning after the default 2-minute reclaim idle threshold and holds the taskId lock.\n2. Redis auto-claims the same pending entry to consumer B.\n3. B cannot acquire the lock, returns normally, and XACKs the shared stream entry.\n4. If A crashes or is killed after B's acknowledgement, there is no pending entry left to reclaim and no retry message. The version can remain in SCANNING, recreating the strand this PR is meant to eliminate.\n\nskipCleanup protects the temp path but does not protect delivery state. Please make lock contention preserve eventual processing—for example, wait for the watchdog lock and re-check isTaskAlreadyProcessed after acquisition, or add an explicit consumer outcome that leaves a contended message pending without acknowledging it. Add a test that drives the auto-claim/lock-contention path and proves the entry is not acknowledged before the active scan reaches a terminal result.\n\nI independently inspected the max-attempt terminal transition, SKIP LOCKED query, soft-delete cleanup, and metadata migration; those address the earlier points. Remote CI/DCO/CLA are green. My local targeted Maven run was blocked only because the available JDK does not support the project's Java 21 --release flag.

FenjuFu:
Thanks for addressing all five review points. I rechecked the latest head (1556665) against #749:

- exhausted retries now reach FAILED and move versions still in SCANNING to SCAN_FAILED;
- FOR UPDATE SKIP LOCKED avoids cross-replica claim collisions;
- the taskId-scoped Redisson lock prevents concurrent duplicate scans without deleting the active scan's temporary files;
- soft deletion removes queued outbox tasks;
- the JSONB metadata migration preserves the original task payload.

DCO, CLA, unit tests, RISC-V builds, and the real-services E2E check are all green. The implementation now satisfies the issue's expected recovery and explicit-failure behavior. I am leaving this as a comment only, not an approval.

XiaoSeS:
已按 review 意见修复并推送到新 head `3ef425a5`。锁竞争时不再正常 return：该分支改为进入统一失败重试路径，并新增回归测试确认重新发布 `retryCount=1` 的消息，避免 reclaim 消费者 ACK 掉唯一投递。

本地 Java 21 验证：skillhub-app 751 passed / 1 skipped；exact-SHA Server 镜像、PostgreSQL、Redis、健康检查和 bootstrap admin 登录均通过。请重新 review；本轮 CI 正在运行。

XiaoSeS (approved):
已验证修复：锁竞争分支进入统一重试路径，新增回归测试覆盖 reclaim/lock contention；Java 21 后端测试、exact-SHA 镜像、运行态、CI 与 E2E 均通过。批准合并。

PR #760, opened 2026-08-26
fix(i18n): keep Russian locale keys in sync

XiaoSeS:
## Summary
- add Russian translations for all keys currently present in English locale
- restore locale key parity so the Web locale test and CI can pass

## Validation
- `pnpm exec vitest run src/i18n/ru-locale.test.ts`
- `pnpm run build`
- English/Russian leaf-key comparison: 0 missing keys

This is a standalone main-branch CI hotfix discovered while validating PR #756; it is not part of #756.


FenjuFu (commented):
Reviewed as a standalone main-branch CI hotfix, not part of #756.

Verified:
- The added keys are the `adminNamespaces.*` block (65 keys) that `en.json` gained with the namespace-admin UI but `ru.json` was missing — the exact gap that fails the leaf-key parity check.
- Parity is confirmed by CI: `Web Build And Test` (which runs `ru-locale.test.ts`) is green, alongside E2E (Real Services), Server Unit Tests, DCO and CLA — all passing.
- Change is scoped to `web/src/i18n/locales/ru.json` only; translations read naturally and use the same ICU placeholders as `en.json`.

Low-risk locale-only fix that unblocks the main-branch locale gate. LGTM.

PR #770, opened 2026-08-28
feat(namespace): provision personal namespace with stable random slug

XiaoSeS:
Maintainer-scoped implementation for #712/#711.

- Provision only newly activated accounts after commit.
- Use `personal-${random}` as the default slug template so usernames and languages never affect technical identifiers.
- Use `${username}-个人空间` as the default display name.
- Keep slug stable and globally unique; display name remains editable.
- No existing-user backfill or default-membership redesign in this PR.

Validation: domain provisioning and naming tests pass.


XiaoSeS:
按维护者流程改回原始 PR 处理；此临时维护者 PR 关闭，后续修复将回到对应原 PR。

PR #776, opened 2026-08-29
fix(web): preserve HTTPS in sub-path redirects

XiaoSeS:
## What

- emit a relative `Location: /skillhub/` for the bare sub-path redirect
- add a regression assertion for the redirect target, not only its status code

## Why

The latest main edge deployment behind the Hong Kong TLS proxy returned:

```text
GET https://skill.xf-yun.com.cn/skillhub
301 Location: http://skill.xf-yun.com.cn/skillhub/
```

Nginx generated an absolute redirect using the Web container's internal HTTP scheme. A relative redirect preserves the browser's external HTTPS scheme and host.

## Testing

- `sh -n web/docker-entrypoint.d/20-base-path.sh`
- `bash -n scripts/tests/web-base-path-nginx-smoke-test.sh`
- `bash scripts/tests/web-base-path-nginx-smoke-test.sh`
- `bash scripts/tests/nginx-forwarded-proto-test.sh`
- `bash scripts/tests/web-base-path-routing-test.sh`
- `bash scripts/tests/runtime-secret-test.sh`
- `bash scripts/tests/validate-release-config-test.sh`
- `git diff --check`

After CI, the fix will be deployed as a new main edge image and rechecked through the real HTTPS domain before release approval.

### DIFF
--- a/scripts/tests/web-base-path-nginx-smoke-test.sh
+++ b/scripts/tests/web-base-path-nginx-smoke-test.sh
@@ -86,6 +86,11 @@ if [ "$code" != '301' ]; then
   echo "bare prefix must 301-redirect, got: $code" >&2
   exit 1
 fi
+location=$(curl -s -o /dev/null -D - "$base/skillhub" | awk 'tolower($1) == "location:" { print $2 }' | tr -d '\r')
+if [ "$location" != '/skillhub/' ]; then
+  echo "bare prefix redirect must stay relative to preserve an upstream HTTPS scheme, got: $location" >&2
+  exit 1
+fi
 
 docker rm -f "$name" >/dev/null 2>&1 || true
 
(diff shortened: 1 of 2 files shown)


FenjuFu (commented):
Verified — the fix is correct.

Adding `absolute_redirect off;` to the `location = /skillhub` block makes nginx emit a relative `Location: /skillhub/` instead of reconstructing an absolute URL from the container's internal `http` scheme. Behind the Hong Kong TLS proxy the browser then keeps its external `https://` scheme and host, which is exactly the reported failure (`301 Location: http://…/skillhub/`).

The new smoke-test assertion pins the exact `Location: /skillhub/` target rather than only the 301 status, so a future regression to an absolute redirect would be caught. Scope is minimal (entrypoint template + test). LGTM.

Issue #516, opened 2026-06-12
[Feature] Import & sync skills from community skill repos (agent-skills style layouts)


FenjuFu:
## Motivation

Community skill libraries are exploding right now — `addyosmani/agent-skills` (51k+ stars) established a de-facto directory convention for production-grade coding skills, `obra/superpowers` (220k+ stars) ships a full skill framework with templates, and single-purpose skills like `mvanhorn/last30days-skill` go viral within days. A huge amount of high-quality skill content now lives in plain Git repos following these community conventions (SKILL.md / skill directory layouts).

Right now, bringing such a community skill into skillhub means manually repackaging it. For teams that want skillhub as their single registry (with versioning, RBAC and audit on top), this is the main friction point: the content exists, but the import path doesn't.

## Proposed feature

An **importer that syncs skills from external Git repos in community formats** into skillhub packages:

1. **Source registration**: point skillhub at a public/internal Git repo URL (+ optional subdirectory), e.g. an `agent-skills`-style collection.
2. **Format mapping**: parse the community layout (skill directory with SKILL.md / frontmatter metadata) and map it to a skillhub package — name, description, version derived from git tag or commit.
3. **Sync policy**: manual one-shot import to start; optionally a scheduled re-sync that creates a new package version when upstream changes (never auto-publishes — goes through the normal review/approval flow).
4. **Governance inherited for free**: imported skills become regular skillhub packages, so RBAC, audit logs, and version pinning all apply — which is exactly what plain Git repos can't provide.

A CLI subcommand (`skillhub import <git-url>`) would already cover most of the value; UI can come later.

## Why it fits this repo

skillhub's positioning is the enterprise registry layer on top of agent skills. The community has now converged on de-facto skill formats faster than any spec process — meeting them with a first-class import path turns every popular community skill repo into potential skillhub content, instead of competing with them on authoring.

## References

- https://github.com/addyosmani/agent-skills
- https://github.com/obra/superpowers
- https://github.com/mvanhorn/last30days-skill

brandonfang06:
Adding a concrete repro and scope note.

`obra/superpowers` uses a plugin-style layout like this:

```text
superpowers/
├── .claude-plugin/plugin.json
├── hooks/hooks.json
└── skills/
    ├── using-superpowers/SKILL.md
    ├── brainstorming/SKILL.md
    ├── test-driven-development/SKILL.md
    └── ... 14 skills total
```

[...]

PR #781, opened 2026-08-29
feat(cli): filter sync push skill batches

FenjuFu:
﻿## Summary

- Add repeatable `sync push --all --include <skill>` filters for publishing a reviewed subset from a large local Agent Skills collection.
- Preserve the requested include order and deduplicate repeated names.
- Validate every requested directory name and existence before any package validation request, preventing partial batches caused by a typo.
- Reject `--include` without `--all` and on non-`push` sync actions instead of silently ignoring it.
- Document the command in CLI help, README, command reference, and changelog.

## Why

Large standards-based collections such as `K-Dense-AI/scientific-agent-skills` contain 160+ immediate skill directories. Operators often want a governed topical subset (for example `scanpy`, `rdkit`, and `literature-review`) without copying directories into a temporary workspace or validating the entire collection against registry request limits.

## Validation

```text
bun test test/integration/sync-command.test.ts --timeout 10000
# 11 passed

bun test test/integration/help-command.test.ts
# 9 passed

bun run typecheck
bun run lint
bun run build
# all passed
```

`bun test` for the whole CLI tree was attempted. On this Windows runner it hung in the existing `update --check` network test and was terminated; the relevant sync/help suites pass, and all new tests pass under Bun's default 5-second threshold. Remote Linux CI remains authoritative for the complete suite.

## Behavior examples

```bash
skillhub sync push --all \
  --include scanpy \
  --include rdkit \
  --namespace research \
  --dry-run
```

A missing requested directory exits before the first registry validation call. Existing `sync push --all` behavior is unchanged when no `--include` is supplied.

## Risk

- CLI-only additive option; no server/API/database/UI changes.
- No overwrite semantics changed.
- Rollback is the single commit.

## Related

- Refs #516
- K-Dense governed publishing guide: https://github.com/K-Dense-AI/scientific-agent-skills/pull/241
- Astron repository import proposal: https://github.com/iflytek/astron-agent/issues/1658

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable CLI behavior changes are documented in this file.
 
 ## Unreleased
 
+### Added
+
+- Add repeatable `sync push --all --include <skill>` filters so operators can
+  validate and publish a reviewed subset from a large local Skill collection.
+  Invalid or missing directory names fail before any package validation request.
+
 ### Fixed
 
 - Resolve `namespace/slug`, `@namespace/slug`, and `namespace--slug`
--- a/cli/README.md
+++ b/cli/README.md
@@ -263,11 +263,19 @@ skillhub sync pull --namespace team-a --prune
 # Validate and upload every local skill for review
 skillhub sync push --all --namespace team-a --dry-run
 skillhub sync push --all --namespace team-a --submit-review
+
+# Validate only a reviewed subset from a larger collection
+skillhub sync push --all --include scanpy --include rdkit --namespace team-a --dry-run
 ```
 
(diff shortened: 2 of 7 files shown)


XiaoSeS:
Thanks for the contribution and for documenting the large-workspace use case. After maintainer review, we are closing this PR because it is not tied to an accepted standalone SkillHub issue and does not implement the import/discovery workflow tracked by #516. The current CLI can publish a selected subset by invoking `skillhub publish <path>` for each chosen skill, so we do not want to add the extra `sync push --all --include` command surface at this time. We can revisit batch-selection support if a concrete requirement remains after the #516 import workflow is designed.

FenjuFu:
Thanks for the clear maintainer decision. Understood: the current per-skill `skillhub publish <path>` flow is sufficient for selected subsets, and a new `sync push --all --include` surface should wait for an accepted requirement after #516's import/discovery design. I will not reopen or rework this PR without that product scope.

PR #831, opened 2026-09-08
fix(publish): ignore runtime secret expressions

### LINKED ISSUE #827: [Bug] Regression: token variable assignments are again flagged as leaked secrets
## Summary

This is a regression of #234. The false-positive detection was fixed by #253 (`0497f865`), but #288 (`2def67b0`) removed the literal/expression classification while introducing the warning-confirmation flow and restored the original broad regular expression.

The current `main` branch again treats ordinary identifiers and function-call results assigned to variables such as `access_token`, `refresh_token`, or `client_secret` as hardcoded credentials.

Related issues and pull requests:

- Original bug: #234
- Original fix: #253
- Change that reintroduced the old matcher: #288

## Steps To Reproduce

1. Create a Skill containing a Python file with ordinary runtime credential handling:

```python
refresh_token = token_response.get("refresh_token")
client_secret = client_secret
self._client_secret = client_secret
access_token = ensure_valid_access_token(...)
headers = build_headers(access_token=access_token)
```

2. Run pre-publish validation:

```bash
skillhub publish ./example-skill --dry-run --json
```

3. The response reports secret/token warnings for these lines and returns `valid: false`, even though none of the lines contains a credential literal.

The current generic rule is:

```regex
(?i)(api[_-]?key|access[_-]?key|secret|password|token)\s*[:=]\s*['\"]?([A-Za-z0-9_\-]{12,})
```

For example, in `refresh_token = token_response.get(...)`, it can match the `token` suffix on the left and capture `token_response` as the alleged secret value. Similarly, it captures `client_secret` or `ensure_valid_access_token` from other normal assignments.

`2def67b0` changed findings from blocking errors to confirmable warnings, but it also deleted the `extractMatchedValue`, `looksLikeExpression`, identifier, quoted-literal, and bare-literal handling introduced by #253. Those are separate concerns: warning severity should not require reverting value classification.

## Expected Behavior

The validator should:

- allow identifiers, function calls, attribute access, and other runtime expressions assigned to sensitive variable names;
- continue warning on actual hardcoded credential literals and high-confidence provider token formats;
- preserve the warning and `confirmWarnings` behavior introduced by #288.

A suitable fix would restore the value-extraction and expression-classification behavior from #253 while returning `ValidationResult.warn(...)` rather than restoring the former blocking result.

Regression tests should cover all of these non-secret assignments, not only a variable literally named `token`, while retaining positive tests for quoted and provider-specific credential values.

## Environment

- Upstream branch: current `main`
- Current source inspected at: `19cc56be9e0d4aab8418a0182e4c05cfc19e9e42`
- CLI: `@astron-team/skillhub@0.1.11`
- Deployment: self-hosted SkillHub

## API Contract Impact

No API contract change is required. The affected endpoint is:

```text
POST /api/cli/v1/skills/{namespace}/publish/validate
```

The existing response shape can remain unchanged. Correcting the matcher should remove false-positive warning entries and allow `valid: true` when no real credential is present.

## Logs Or Screenshots

Representative response shape:

```json
{
  "valid": false,
  "errors": [],
  "warnings": [
    "scripts/example.py line 1 contains a value that looks like a secret or token. Replace real credentials with placeholders before publishing."
  ]
}
```

FenjuFu:
## Summary

- restore the literal-versus-expression classification introduced by #253 and accidentally removed by #288
- ignore identifiers, function calls, attribute/index access, environment expansion, and nested runtime arguments assigned to sensitive variable names
- preserve the current warning-only result and `confirmWarnings` flow from #288
- continue warning on quoted and bare credential literals and on high-confidence provider token formats wherever they appear
- cover every representative false positive from #827 plus positive hardcoded-secret cases

Closes #827.

## Why

The broad assignment regex currently captures values such as `token_response`, `client_secret`, and `ensure_valid_access_token` as if they were credential literals. Severity and value classification are separate concerns: this change restores classification while retaining confirmable warnings for actual literals.

## Validation

- `git diff --check` — passes
- regression tests added for identifiers, calls, attributes, indexing, nested keyword arguments, quoted literals with comments, bare literals, and embedded GitHub tokens
- local Maven execution could not reach JUnit because this host only has JDK 17 while the repository compiles with `--release 21`; GitHub Java 21 CI is the authoritative test run
- no API contract or generated files changed

### DIFF
--- a/server/skillhub-domain/src/main/java/com/iflytek/skillhub/domain/skill/validation/BasicPrePublishValidator.java
+++ b/server/skillhub-domain/src/main/java/com/iflytek/skillhub/domain/skill/validation/BasicPrePublishValidator.java
@@ -17,6 +17,12 @@ import java.util.regex.Pattern;
 @Component
 public class BasicPrePublishValidator implements PrePublishValidator {
 
+    private static final Pattern ASSIGNMENT_WITH_SENSITIVE_KEY = Pattern.compile(
+            "(?i)(api[_-]?key|access[_-]?key|secret|password|token)\\s*[:=]\\s*(.+)$"
+    );
+    private static final Pattern QUOTED_LITERAL = Pattern.compile("^(['\"])(.*)\\1$");
+    private static final Pattern IDENTIFIER = Pattern.compile("[A-Za-z_][A-Za-z0-9_]*");
+    private static final Pattern BARE_LITERAL = Pattern.compile("[A-Za-z0-9_\\-]{12,}");
     private static final Set<String> PLACEHOLDER_MARKERS = Set.of(
             "your", "example", "sample", "placeholder", "changeme", "replace", "dummy",
             "mock", "test", "fake", "todo", "xxx", "redacted");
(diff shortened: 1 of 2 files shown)


XiaoSeS:
Closing as superseded by #830, which has already merged and closes #827 with the credential-expression classification, positive hardcoded-secret coverage, linear-time parsing, and full Java 21 CI validation. Thank you for the alternative implementation and regression coverage.

PR #862, opened 2026-09-13
fix: resolve i18n language switching bug in LanguageSwitcher component

hssuryawanshib23-afk:
Fixed the language switcher in the web interface so users can switch to English properly.

The problem was that the language switcher was hardcoded to default to Chinese ('zh')
instead of following the app's language settings, which meant clicking English had no effect.

Changes made:
- Fixed the hardcoded fallback so it now respects the app's language configuration
- Added a safety check to make sure the switcher updates when language changes

This resolves the issue where switching to English in the language selector had no effect.


hssuryawanshib23-afk:
Please let me know if you'd like any changes or have any suggestions for improvement.

XiaoSeS:
Thanks for the contribution. I am closing this PR rather than merging it for two concrete reasons:

1. The required DCO check is failing on the current commit.
2. The implementation adds an empty `useEffect` that has no runtime effect. The current `LanguageSwitcher` already derives the displayed language from `i18n.language`, and `useTranslation()` re-renders consumers after `changeLanguage`; the hardcoded `zh` fallback only applies when no language is available and does not by itself explain English clicks having no effect.

If this bug is still reproducible, please reopen with a concrete reproduction path and a focused test that fails on `main`, then fix the actual failing path without adding no-op effects.

XiaoSeS:
补充验证：我刚才在最新 `origin/main` 上增加了一个临时真实交互单测，用真实 `i18next.createInstance()` + `I18nextProvider` 渲染 `LanguageSwitcher`，初始语言 `zh`，点击菜单里的 `English` 后断言按钮显示变成 `English` 且 `i18n.language === "en"`。

执行命令：

```bash
pnpm --dir web exec vitest run src/shared/components/language-switcher.test.ts
```

结果：1 个测试文件、2 个用例通过。

因此当前 `main` 上未复现“点击 English 无效”的问题。关闭原因不是只因为实现风格，而是：当前 PR 的 DCO 失败、实现包含无效的空 `useEffect`，并且它声称修复的问题在最新 main 上没有复现。若后续仍能复现，请提供具体入口、浏览器/语言检测条件、localStorage 状态和失败测试。

XiaoSeS:
Thanks for reopening this. I re-checked the report and found the real issue worth fixing: the switcher could still default its label to Chinese when i18next had not resolved a language yet, while the app-level i18n fallback is English.\n\nI did not merge this PR because the added empty `useEffect` has no runtime effect. Instead, I fixed the actual fallback behavior in #872:\n\n- LanguageSwitcher now prefers `i18n.resolvedLanguage`, then `i18n.language`, then English.\n- Unknown/unresolved language state falls back to English, matching `web/src/i18n/config.ts`.\n- Added jsdom interaction coverage for unresolved language, resolved-language precedence, and selecting English.\n\nValidation for #872 passed:\n\n- `pnpm --dir web exec vitest run src/shared/components/language-switcher.test.tsx`\n- `make typecheck-web`\n- `make lint-web`\n- GitHub CI: DCO, CLA, Docs, Server Unit Tests, Web Build And Test, Dependency Review, RISC-V builds, and E2E Real Services.\n\n#872 has been merged, so I am closing this PR in favor of that fix.

XiaoSeS:
Closed in favor of merged PR #872, which fixes the validated LanguageSwitcher fallback issue with tests.

PR #879, opened 2026-09-18
fix(scanner): harden Scanner 2.1 integration

### LINKED ISSUE #868: [Feature] Upgrade cisco-ai-skill-scanner to 2.x to improve secret-leak and script scanning coverage
### Problem

**Background / Current State**
SkillHub currently uses cisco-ai-skill-scanner 1.0.2 for Skill security scanning. This version has limited scanning scope and extensibility, leaving notable security blind spots:

- Limited coverage of script and file types.
- 
- Cannot detect sensitive information leaks in JavaScript files.
- 
- Cannot scan .env, config, or similar configuration files for leaked keys/credentials.
- 
- Limited extensibility, making it hard to cover new file types and detection rules.

As a result, some Skills may ship with hardcoded secrets, tokens, or sensitive config that the current scanner fails to flag.

### Proposed Solution

Newer releases of cisco-ai-skill-scanner (e.g. 2.1.0) already support these capabilities, including:

- JavaScript script scanning.
- 
- Secret-leak detection in .env, config, and similar files.
- 
- Broader file-type coverage and better extensibility.
- 
- Overall improved security posture.

It would be great to upgrade SkillHub's dependency from 1.0.2 to 2.x (e.g. 2.1.0) to improve scan coverage and security.

### Alternatives Considered

_No response_

### Impact

_No response_

### Contract Or SDK Impact

_No response_

dongmucat:
## What
- harden the SkillHub integration with Cisco AI Skill Scanner 2.1
- keep AI Defense credentials in `X-AIDefense-Key` headers only
- normalize the documented `azure` provider alias to Scanner 2.1 `azure-openai`
- redact credentials from structured findings without mutating safe text
- isolate and recreate scanner runtime temporary storage across container restarts
- add regression tests, Docker contract coverage, and rollout documentation

## Why
Closes #868. The Scanner 2.1 upgrade needs a safe request contract, bounded runtime behavior, and protection against credential leakage and stale temporary files.

## How
- Scanner wrapper uses the pinned 2.1 credential regexes while preserving multiline and long finding text.
- Java scanner options validate providers and send AI Defense keys only via headers.
- The scanner-owned `0700` temporary root is recreated before upstream import, allowing hard-timeout restarts to reclaim abandoned files without touching global `/tmp`.
- Mixed Scanner 1.0.2/2.1 rollout guidance keeps AI Defense disabled until old instances are drained.

## Testing
- `make test-backend-app` — 892 tests passed, 0 failures, 0 errors, 1 skipped
- Python Scanner tests — 12 passed
- `make typecheck-web` — passed
- `make lint-web` — passed
- Scanner 2.1 Docker contract — passed, including 51 MiB upload and same-container restart cleanup
- `shellcheck`, `bash -n`, and `git diff --check` — passed

## Impact
No frontend API contract changes or database migrations. During a mixed-version rollout, keep AI Defense disabled until old Scanner/Server instances are drained as documented.


FenjuFu (approved):
Checked this against #868: the Scanner image now pins `cisco-ai-skill-scanner==2.1.0`, which brings in the JS and config/secret coverage the issue asks for. The 1.0.2 base-URL backport is dropped, the contract test verifies the reported version, and the rolling-upgrade ordering is documented in both FAQs. Keeping AI Defense keys header-only and redacting credentials from findings are good hardening on top. All checks are green, including both scanner contract architectures. LGTM.

scanner/skillhub_scanner_app.py
```diff
@@ -5,32 +5,92 @@
...
+    if isinstance(value, list):
+        return [_redact_supported_tokens(item) for item in value]
+    if isinstance(value, dict):
+        return {key: _redact_supported_tokens(item) for key, item in value.items()}
+    return value
+
+
+def _install_scan_response_redaction() -> None:
+    """Redact supported token forms before FastAPI serializes scan findings."""
+    for route in _upstream_router.router.routes:
```
dongmucat, line 79: [P1] Install redaction on the mounted FastAPI routes. `skill_scanner.api.api` calls `app.include_router(api_router)` before this module imports `app`; FastAPI creates new `APIRoute` and `Dependant` objects there. Mutating `_upstream_router.router.routes` therefore leaves the handlers in `app.routes` unchanged, so HTTP `/scan` and `/scan-upload` responses bypass `redacting_endpoint` and can persist or return unredacted finding values. I reproduced the route behavior with FastAPI: `mounted is original == False`, and replacing the original `dependant.call` still invokes the unwrapped mounted handler. Please patch the mounted routes (or wrap before inclusion) and add an HTTP-level canary test. The current Python tests call the original router route directly, while the Docker contract only checks canaries that the upstream scanner may already mask.

